PR #20, opened 2020-05-29
fix: TN-1251 Fix pull/push of semantic model;

meritonhusaj:
Fixed pull of semantic model to be yaml and not json;
Fixed push to also activate that semantic model;


ErmalFerati (approved):
It'd be very cool if we added a --publish and --activate parameter later on

PR #57, opened 2021-04-15
feat TN Adapt build action to publish to public registry

promeris:
#### Description

- Add README and LICENSE file to build
- Add step to publish to public npm registry

#### Checklist

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -14,6 +14,14 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Setup Node with Github Registry
+        uses: actions/setup-node@v1
+        with:
+          node-version: '10.17.0'
+          registry-url: https://npm.pkg.github.com/celonis
+          scope: '@celonis'
+          token: ${{ secrets.GITHUB_TOKEN }}
+
       - name: Yarn Install
         run: yarn install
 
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-publish.yml
```diff
@@ -14,19 +14,43 @@ jobs:
...
         run: yarn install
 
       - name: Yarn Build
         run: yarn build
 
-      - name: Publish library
-        run: cd dist/ && npm publish --access public
+      - name: Move files to build
+        run: |
+          cp package.json dist/package.json
```
ErmalFerati, line 25: Maybe we should do this using the build script instead. That way even if change the build workflows then it'll still work.
We already do this for `package.json`.

GencBlakqoriP9 (dismissed):

.github/workflows/build-and-publish.yml
GencBlakqoriP9: I was thinking the same actually.

GencBlakqoriP9:
For the future I would recommend that we have three different jobs instead of doing the pushing in one JOB:
1) One for building the lib and push it to the current release
2) Push lib to github
3) push lib to npmjs

The last two can run in paralel. IMO it's a bit cleaner like this + these steps are independent.

.github/workflows/build-and-publish.yml
promeris: I got a message that someone was trying to build the project on windows and it wasn't working because `cp` is not a command in cmd, that's why i moved these to a step in the build.

.github/workflows/build-and-publish.yml
```diff
@@ -14,19 +14,43 @@ jobs:
...
+          node-version: '10.17.0'
+          registry-url: https://registry.npmjs.org/
+
+      - name: Publish to Npm Registry
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+          CI: true
+        run: |
+          cd dist/
+          npm config set //registry.npmjs.org/:_authToken=$NODE_AUTH_TOKEN
```
ErmalFerati, line 41: We can set this in the `Setup publish to Npm Registry` step instead.

I think something like this should do the trick (we might have to do it twice since you can't specify a registry in the publish command):
```
- uses: actions/setup-node@v1
  with:
    node-version: 10.x
    npm-registry:
      - registry: 'https://npm.pkg.github.com'
        scope: '@celonis'
        token: '${{secrets.GITHUB_TOKEN}}'
      - registry: 'https://registry.npmjs.org'
        token: '${{secrets.NODE_AUTH_TOKEN}}'
```

promeris: I checked the `setup-node` implementation and the [registry-url](https://github.com/actions/setup-node/blob/v2.1.5/src/main.ts#L42) input is only a string so it wouldn't work.

meritonhusaj (dismissed):

meritonhusaj:
> For the future I would recommend that we have three different jobs instead of doing the pushing in one JOB:
> 
> 1. One for building the lib and push it to the current release
> 2. Push lib to github
> 3. push lib to npmjs
> 
> The last two can run in parallel. IMO it's a bit cleaner like this + these steps run independent.

Do we need to have the same version in both? I'm just curious if it's on public registry, why would we need the same version also on our private one.

promeris:
> > For the future I would recommend that we have three different jobs instead of doing the pushing in one JOB:
> > 
> > 1. One for building the lib and push it to the current release
> > 2. Push lib to github
> > 3. push lib to npmjs
> > 
> > The last two can run in parallel. IMO it's a bit cleaner like this + these steps run independent.
> 
> Do we need to have the same version in both? I'm just curious if it's on public registry, why would we need the same version also on our private one.

@meritonhusaj Unfortunately organizational members can't install the latest published versions in npmjs if they have the scoped registry configuration in their npmrc configuration, because npm will search for scoped pacakges only on the defined registry and not fallback to the public one.

meritonhusaj (approved):

ErmalFerati (approved):

PR #58, opened 2021-04-15
pass statusCode correctly to handleBadRequest(...)

bperquku:
#### Description

Sometimes when pushin package got error: 
     error:   Backend responded with status code [object Object]

Fixed by passing statusCode correctly

#### Checklist


src/services/http-client.service.ts
```diff
@@ -94,7 +94,7 @@ export class HttpClientService {
     // tslint:disable-next-line:typedef
     private handleResponse(res: Response, resolve, reject): void {
         if (this.checkBadRequest(res.statusCode)) {
-            this.handleBadRequest(res, res.body, reject);
+            this.handleBadRequest(res.statusCode, res.body, reject);
```
ErmalFerati, line 97: It seems that the same thing happens in line 35, could you also handle that case?

promeris:
Can you also please bump the patch version in package.json to 0.0.51, because the bump action is not working correctly at the moment.

promeris (dismissed):

qmucolli (dismissed):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@celonis/content-cli",
-  "version": "0.1.0",
+  "version": "0.1.1",
```
qmucolli, line 3: You probably need to bump this once more as PR's have already been merged to master and this won't get published

qmucolli (approved):

PR #61, opened 2021-04-22
feat TN-1962 added studio spaces support for package import

qmucolli:
This PR contains the new changes required for the introduction of Studio Spaces.

Affected actions:
- push package
- push packages

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celonis/content-cli",
-  "version": "0.0.51",
+  "version": "0.0.52",
   "description": "CLI Tool to help manage content in Celonis EMS",
   "main": "content-cli.js",
   "bin": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celonis/content-cli",
-  "version": "0.0.52",
+  "version": "0.1.51",
   "description": "CLI Tool to help manage content in Celonis EMS",
   "main": "content-cli.js",
   "bin": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celonis/content-cli",
-  "version": "0.1.51",
+  "version": "0.1.0",
   "description": "CLI Tool to help manage content in Celonis EMS",
   "main": "content-cli.js",
   "bin": {
--- a/src/commands/package.command.ts
+++ b/src/commands/package.command.ts
@@ -25,8 +25,8 @@ export class PackageCommand {
         );
     }
 
-    public async pushPackages(profile: string): Promise<void> {
-        await this.contentService.batchPush(profile, this.packageManagerFactory.createManagers());
+    public async pushPackages(profile: string, spaceKey: string): Promise<void> {
+        await this.contentService.batchPush(profile, this.packageManagerFactory.createManagers(spaceKey));
     }
 
     public async listPackages(profile: string): Promise<void> {
--- a/src/commands/package.command.ts
+++ b/src/commands/package.command.ts
@@ -6,13 +6,22 @@ export class PackageCommand {
     private packageManagerFactory = new PackageManagerFactory();
 
     public async pullPackage(profile: string, key: string, store: boolean, newKey: string): Promise<void> {
-        await this.contentService.pullFile(profile, this.packageManagerFactory.createManager(key, null, store, newKey));
+        await this.contentService.pullFile(
+            profile,
+            this.packageManagerFactory.createManager(key, null, null, store, newKey)
+        );
     }
 
-    public async pushPackage(profile: string, fileName: string, newKey: string, overwrite: boolean): Promise<void> {
+    public async pushPackage(
+        profile: string,
+        spaceId: string,
+        fileName: string,
+        newKey: string,
+        overwrite: boolean
+    ): Promise<void> {
         await this.contentService.push(
             profile,
-            this.packageManagerFactory.createManager(null, fileName, false, newKey, overwrite)
+            this.packageManagerFactory.createManager(null, spaceId, fileName, false, newKey, overwrite)
         );
     }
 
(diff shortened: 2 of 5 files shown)


src/content-cli-push.ts
```diff
@@ -139,12 +139,13 @@ class Push {
         program
             .command("package")
             .description("Command to push a package to Studio")
+            .option("--spaceId <spaceId>", "Id of the destination space")
```
meritonhusaj, line 142: I'd call this spaceKey. Wouldn't prefer to introduce a new terminology to the end users. They are familiar with keys for packages, and we also show it as `Key` in menu of the Spaces for the same reason

src/content/manager/package.manager.ts
```diff
@@ -91,26 +100,43 @@ export class PackageManager extends BaseManager {
...
+
+        if (hasErrors) {
+            process.exit();
+        }
+    }
+
+    private getPushUrlWithParams(pushUrl: string): string {
+        let pushUrlWithParams = `${pushUrl}?spaceId=${this.spaceKey}&`;
+        if (this.newKey) {
+            pushUrlWithParams = `${pushUrlWithParams}newKey=${this.newKey}&`;
```
promeris, line 135: nit: Let's return on both if statements because we don't want to have both the newKey and overwrite request params. We also validate this case on the above method.

promeris (commented):
I think we should also increment the minor version

qmucolli:
> I think we should also increment the minor version

Why do you think so? The main reason we'd increment it, is that the users of that lib would not have breaking changes. I doubt that someone is using our version right now as it was increased during the last release and we didn't release this one yet.

promeris:
> > I think we should also increment the minor version
> 
> Why do you think so? The main reason we'd increment it, is that the users of that lib would not have breaking changes. I doubt that someone is using our version right now as it was increased during the last release and we didn't release this one yet.

I don't have a strong opinion on this but I think that because we are introducing a breaking change to the API, this should also be reflected to the tool that is using it by incrementing the version and communicate that these commands will work after version x.y with the new Studio release.

qmucolli:
> > > I think we should also increment the minor version
> > 
> > 
> > Why do you think so? The main reason we'd increment it, is that the users of that lib would not have breaking changes. I doubt that someone is using our version right now as it was increased during the last release and we didn't release this one yet.
> 
> I don't have a strong opinion on this but I think that because we are introducing a breaking change to the API, this should also be reflected to the tool that is using it by incrementing the version and communicate that these commands will work after version x.y with the new Studio release.

Ups sorry, you're actually right. I didn't notice you commented on the cli repo, I thought you commented on the one in package-manager 😅  I'll fix it accordingly 👍

src/commands/package.command.ts
```diff
@@ -6,18 +6,27 @@ export class PackageCommand {
...
+        await this.contentService.pullFile(
+            profile,
+            this.packageManagerFactory.createManager(key, null, null, store, newKey)
+        );
     }
 
-    public async pushPackage(profile: string, fileName: string, newKey: string, overwrite: boolean): Promise<void> {
+    public async pushPackage(
+        profile: string,
+        spaceId: string,
```
GencBlakqoriP9, line 17: Why di we use here a `spaceId` while in the push commands we're use `spaceKey`?

src/content/manager/package.manager.ts
```diff
@@ -91,26 +100,43 @@ export class PackageManager extends BaseManager {
...
-                logger.error(
-                    "You cannot overwrite a package and set a new key at the same time. Please use only one of the options."
-                );
-                process.exit();
-            } else {
-                pushUrl = `${pushUrl}?overwrite=${this.overwrite}`;
-            }
-        }
-        return pushUrl;
+        this.validateOptions();
```
GencBlakqoriP9, line 103: IMO, this is not the place to validate the options, since the method's responsibility is to build the push url. I think such validation things should be handled in `package.command.ts`, before the factories are even created.

GencBlakqoriP9:
> > > > I think we should also increment the minor version
> > > 
> > > 
> > > Why do you think so? The main reason we'd increment it, is that the users of that lib would not have breaking changes. I doubt that someone is using our version right now as it was increased during the last release and we didn't release this one yet.
> > 
> > 
> > I don't have a strong opinion on this but I think that because we are introducing a breaking change to the API, this should also be reflected to the tool that is using it by incrementing the version and communicate that these commands will work after version x.y with the new Studio release.
> 
> Ups sorry, you're actually right. I didn't notice you commented on the cli repo, I thought you commented on the one in package-manager 😅 I'll fix it accordingly 👍

@qmucolli you actually bumped the patch version. The new version should be `0.1.0`.

src/commands/package.command.ts
qmucolli: Ah, I forgot it there, It's just the parameter name I'll update it 👍

src/content/manager/package.manager.ts
qmucolli: I'm don't have a strong opinion about it either. It just feels more natural having it here, as if we store that validation logic inside of the `package.command.ts` it looks a bit odd IMO, as I think that file should have the responsibility of converting the commands to actual logic calls which do the work for them, otherwise I don't think we should have that at all. Ideally we should have some validator file maybe that does that throughout the app 🤔

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@celonis/content-cli",
-  "version": "0.0.52",
+  "version": "0.1.51",
```
GencBlakqoriP9, line 3: This should be `0.1.0` instead of `0.1.0`

GencBlakqoriP9 (changes requested):

meritonhusaj (dismissed):

meritonhusaj (approved):

PR #62, opened 2021-05-05
CPMBR-2980 Add extra params to ctp push command

karmann-dm:
##### Motivation

According to the new business requirements regarding the CPM to EMS migration we decided to add a possibility to control whether user wants to import all data models into several different pools or into one. 

##### Actual changes

Added following `push ctp` command options:
1. `--globalPoolName <name>` indicates whether user wants to import all data models from CTP file **inside one newly created data pool**
2. `--existingPoolId <ID>` indicates whether user wants to import all data models from CTP file **inside one already existing data pool**

#### Checklist

### DIFF
--- a/src/content-cli-push.ts
+++ b/src/content-cli-push.ts
@@ -32,9 +32,21 @@ class Push {
             .option("-p, --profile <profile>", "Profile which you want to use to push the analysis")
             .option("-a, --pushAnalysis", "Specify this option if you want to push analysis from the CTP file")
             .option("-d, --pushDataModels", "Specify this option if you want to push data models from the CTP file")
-            .option("--globalPool", "", false)
-            .option("--globalPoolName <globalPoolName>", "", null)
-            .option("--existingPoolId <existingPoolId>", "", null)
+            .option(
+                "--globalPool",
+                "Specify this option along with --globalPoolName if you want to push all Data models into one newly created pool",
+                false
+            )
+            .option(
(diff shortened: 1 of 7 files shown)


andrii-kovalenko-celonis (dismissed):

src/content-cli-push.ts
```diff
@@ -32,6 +32,16 @@ class Push {
             .option("-p, --profile <profile>", "Profile which you want to use to push the analysis")
             .option("-a, --pushAnalysis", "Specify this option if you want to push analysis from the CTP file")
             .option("-d, --pushDataModels", "Specify this option if you want to push data models from the CTP file")
+            .option(
+                "--globalPoolName <globalPoolName>",
+                "Specify this option if you want to push all Data models into one newly created pool along with value to set the name of the pool to be created",
+                null
```
GencBlakqoriP9, line 38: Why do we need `null` in these two options?

GencBlakqoriP9 (dismissed):

src/content-cli-push.ts
karmann-dm: @GencBlakqoriP9  It is absolutely valid case if user doesn't specify both of them. And if we don't set the default `null` values - then it will be treated and sent to the backend as `undefined`

meritonhusaj (dismissed):

meritonhusaj (dismissed):

meritonhusaj (approved):

PR #64, opened 2021-05-11
fix: Add spaceKey when pushing CTP file using content-cli.

KhushalGupta:
#### Description

The PR includes the changes to include optional spaceKey while pushing CTP file to Studio. So, here it is an optional field, to provide the spaceKey in the content-cli push ctp command, if you pass spaceKey it pushes the CTP file to that space, else if you don't provide the spaceKey it pushes the CTP file to default space.

#### Checklist

### DIFF
--- a/src/content-cli-push.ts
+++ b/src/content-cli-push.ts
@@ -42,6 +42,11 @@ class Push {
                 "Specify this option if you want to push all Data models into one already existing pool with provided ID",
                 null
             )
+            .option(
+                "-s, --spaceKey <spaceKey>",
+                "The key of the destination space where the analyses from .ctp file will be created.",
+                ""
+            )
             .requiredOption("-f, --file <file>", "The .ctp file you want to push")
             .requiredOption("--password <password>", "The password used for extracting the .ctp file")
             .action(async cmd => {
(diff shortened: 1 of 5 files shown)


src/content-cli-push.ts
```diff
@@ -42,6 +42,11 @@ class Push {
                 "Specify this option if you want to push all Data models into one already existing pool with provided ID",
                 null
             )
+            .option(
```
qmucolli, line 45: What happens if we don't have a spaceKey here?

src/content/manager/ctp.manager.ts
```diff
@@ -4,6 +4,7 @@ import { ManagerConfig } from "../../interfaces/manager-config.interface";
 export abstract class CtpManager extends BaseManager {
     protected _content: any;
     protected _password: string;
+    protected _spaceKey?: string;
```
qmucolli, line 7: Do we need to update the `DOCUMENTATION.md` for these changes?

src/content-cli-push.ts
KhushalGupta: It will select the default space, and if the default space is also not there it id going to throw the exception.

src/content/manager/ctp.manager.ts
KhushalGupta: I am not so sure, as I don't see anything related to this CTP command in DOCUMENTATION.md.

qmucolli (approved):

melhalaby (approved):

PR #69, opened 2021-06-22
Studio manager migration

bavlyshenouda:
#### Description

In preparation for using studio manager for widget administration, widget should be pushed to studio manager instead of the package manager. 

Extra flag is added for activation.

#### Checklist


qmucolli:
Closing because the age of PR and uncertanities of the change.

Issue #68, opened 2021-06-18
Authentication by Application key


kaztakata:
Currently content-cli can access to IBC via user API key only.
Looks like pycelonis.get_celonis method, is it possible to access via Application key too ?

In my environment, this works if I just changed code in profile.validator.js / http-client.service.js from `Bearer ${apiToken}` to `AppKey ${apiToken}`.

promeris:
hey @kaztakata, thanks for reporting the issue. Currently we don't support authentication with an application key in content-cli. You are on the right track with the suggested changes, however replacing `Bearer ${apiToken}` with `AppKey ${apiToken}` in the mentioned files would then only support authentication with application key. The ideal state would be to support both user and application keys for authentication and there are a few changes needed to support it. The following list would summarize the needed changes. 

- Add option to provide application key in the `profile create` command.
- Add support for authenticating via `APPLICATION_TOKEN` environment variable.
- Add logic to conditionally switch between user and application key authentication based on the provided token.

I would say that this task is not complex to implement and would be a good first issue. We would highly appreciate your contribution by creating a pull request with these changes.

kaztakata:
Dear @promeris 
I created pull request https://github.com/celonis/content-cli/pull/71 regarding this issue. It is my first pull request to OSS so please tell me if something incorrect manner.

PR #71, opened 2021-07-11
implement authentication by application key

kaztakata:
#### Description
I made changes to utilize both user API key and Application key regarding https://github.com/celonis/content-cli/issues/68
.

#### Checklist

  - [x] profile create command by user API key
  - [x] profile create command by Application key
  - [x] list package command with profile contains user API key
  - [x] list package command with profile contains Application key
  - [x] list package command with environment variable of user API key
  - [x] list package command with environment variable of Application key
  - [x] pull package command with profile contains Application key
  - [x] push package command with profile contains Application key

### DIFF
--- a/src/commands/profile.command.ts
+++ b/src/commands/profile.command.ts
@@ -12,7 +12,7 @@ export class ProfileCommand {
         profile.name = await QuestionService.ask("Name of the profile: ");
         profile.team = await QuestionService.ask("Your team (please provide the full url): ");
         profile.apiToken = await QuestionService.ask("Your api token: ");
-        await ProfileValidator.validateProfile(profile);
+        profile.authenticationType = await ProfileValidator.validateProfile(profile);
         this.profileService.storeProfile(profile);
         if (setAsDefault) {
             await this.makeDefaultProfile(profile.name);
--- a/src/interfaces/profile.interface.ts
+++ b/src/interfaces/profile.interface.ts
@@ -2,7 +2,7 @@ export interface Profile {
     name: string;
     team: string;
     apiToken: string;
-    authenticationType: string;
+    authenticationType: AuthenticationType;
 }
 
 export type AuthenticationType = "Bearer" | "AppKey";
--- a/src/interfaces/profile.interface.ts
+++ b/src/interfaces/profile.interface.ts
@@ -2,4 +2,5 @@ export interface Profile {
     name: string;
     team: string;
     apiToken: string;
+    authenticationType: string;
 }
--- a/src/interfaces/profile.interface.ts
+++ b/src/interfaces/profile.interface.ts
@@ -4,3 +4,10 @@ export interface Profile {
     apiToken: string;
     authenticationType: string;
 }
+
+export type AuthenticationType = "Bearer" | "AppKey";
+// tslint:disable-next-line:variable-name
+export const AuthenticationType: { [key: string]: AuthenticationType } = {
+    BEARER: "Bearer",
+    APPKEY: "AppKey",
+};
--- a/src/services/http-client.service.ts
+++ b/src/services/http-client.service.ts
@@ -58,20 +58,26 @@ export class HttpClientService {
         });
     }
 
-    public async authenticate(profile: Profile): Promise<any> {
+    public async buildRequestHeadersWithAuthentication(profile: Profile): Promise<any> {
         return new Promise<any>(resolve => {
             const url = profile.team.replace(/\/?$/, "/api/cloud");
             let headers = {
-                authorization: `Bearer ${profile.apiToken}`,
+                authorization: null,
                 "content-type": "application/json",
             };
-            this.checkAuthorization(url, headers)
+            if (profile.authenticationType != null) {
+                headers.authorization = `${profile.authenticationType} ${profile.apiToken}`;
+                resolve(headers);
+            } else {
+                headers.authorization = `Bearer ${profile.apiToken}`;
+            }
+            this.authenticate(url, headers)
                 .then(() => {
                     resolve(headers);
                 })
                 .catch(() => {
                     headers.authorization = `AppKey ${profile.apiToken}`;
-                    this.checkAuthorization(url, headers)
+                    this.authenticate(url, headers)
                         .then(() => {
                             resolve(headers);
                         })
@@ -82,7 +88,7 @@ export class HttpClientService {
         });
     }
 
-    private async checkAuthorization(url: string, headers: Headers): Promise<any> {
+    private async authenticate(url: string, headers: Headers): Promise<any> {
         return new Promise<any>((resolve, reject) => {
             request.get(url, this.makeOptions(headers, null), (err, res) => {
                 this.handleResponse(res, resolve, reject);
--- a/src/services/http-client.service.ts
+++ b/src/services/http-client.service.ts
@@ -5,26 +5,26 @@ import { CoreOptions, Headers, Response } from "request";
 import request = require("request");
 
 export class HttpClientService {
-    public async pushData(url: string, profile: Profile, body: object): Promise<any> {
+    public async pushData(url: string, headers: Headers, body: object): Promise<any> {
         return new Promise<any>((resolve, reject) => {
-            request.post(url, this.makeOptions(profile, body), (err, res) => {
+            request.post(url, this.makeOptions(headers, body), (err, res) => {
                 this.handleResponse(res, resolve, reject);
             });
         });
     }
 
-    public async pullData(url: string, profile: Profile): Promise<any> {
+    public async pullData(url: string, headers: Headers): Promise<any> {
         return new Promise<any>((resolve, reject) => {
-            request.get(url, this.makeOptions(profile, null), (err, res) => {
+            request.get(url, this.makeOptions(headers, null), (err, res) => {
                 this.handleResponse(res, resolve, reject);
             });
         });
     }
 
-    public async pullFileData(url: string, profile: Profile): Promise<any> {
+    public async pullFileData(url: string, headers: Headers): Promise<any> {
         return new Promise<any>((resolve, reject) => {
             request
-                .post(url, this.makeFileDownloadOptions(profile))
+                .post(url, this.makeFileDownloadOptions(headers))
                 .on("response", (response: Response) => {
                     const data: Buffer[] = [];
                     response.on("data", (chunk: Buffer) => {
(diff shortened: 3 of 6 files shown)


src/services/http-client.service.ts
```diff
@@ -42,44 +42,69 @@ export class HttpClientService {
...
         return new Promise<any>((resolve, reject) => {
-            request.get(url, this.makeOptions(profile), (err, res) => {
+            request.get(url, this.makeOptions(headers, null), (err, res) => {
                 this.handleResponse(res, resolve, reject);
             });
         });
     }
 
-    private makeOptions(profile: Profile, body: object = {}): CoreOptions {
+    public async authenticate(profile: Profile): Promise<any> {
```
kaztakata, line 61: add authenticate method to create headers.

src/services/http-client.service.ts
```diff
@@ -42,44 +42,69 @@ export class HttpClientService {
...
+                            resolve(headers);
+                        })
+                        .catch(() => {
+                            resolve(headers);
+                        });
+                });
+        });
+    }
+
+    private async checkAuthorization(url: string, headers: Headers): Promise<any> {
```
kaztakata, line 85: add checkAuthorization method to try to login to IBC.

src/services/http-client.service.ts
```diff
@@ -42,44 +42,69 @@ export class HttpClientService {
...
-    private makeFileDownloadOptions(profile: Profile): object {
+    private makeFileDownloadOptions(headers: Headers): object {
         return {
-            headers: this.buildRequestHeadersWithAuthentication(profile.apiToken),
+            headers: headers,
             responseType: "binary",
         };
     }
 
-    private buildRequestHeadersWithAuthentication(apiToken: string): Headers {
```
kaztakata, line 76: delete buildRequestHeadersWithAuthentication method and add authenticate method instead.

src/validators/profile.validator.ts
```diff
@@ -35,8 +34,15 @@ export class ProfileValidator {
                     reject();
                 }
                 if (res.statusCode >= 400 || body.teamDomain == null) {
-                    logger.error(new FatalError("The provided team or api key is wrong."));
-                    reject();
+                    options.headers.authorization = `AppKey ${profile.apiToken}`;
```
kaztakata, line 37: if login with Bearer failed, change to AppKey and try to login again.

src/services/http-client.service.ts
promeris: I really like this idea of making a simple request to check if the token should be used with the `Bearer` or `AppKey` prefix, however I think that making this decision before each command/http call is too much when using a profile. I would suggest adding a prefix property in the `Profile` interface and setting it to `Bearer` or `AppKey` when validating [here](https://github.com/celonis/content-cli/pull/71/files#diff-c8416a6cab692c74a7a6cd69256d4f2a0c9a96c7ac67628553eec415c5a68508R36), depending on the validation result. Then you could use the prefix from the profile in this class to build the Authorization header accordingly.

You also need to do a similar approach to this `authenticate` method to create the profile prefix [here](https://github.com/celonis/content-cli/blob/5ba40389fc62556a7530a9fefebf3ba71ab19ed6/src/services/profile.service.ts#L71), so command calls with env variables also work. The following command is an example of using env variables instead of profiles.
```
API_TOKEN=<some-token> TEAM_URL=<some-team-url> content-cli push package
```

kaztakata: @promeris 
Thanks for your check. 
Regarding authentication by Profile option I got your point.

On the other hand regarding authentication by env variable, still I am confused.
If I understand your comment correctly, I should create new env variable AUTHORIZATION to set 'Bearer' or 'AppKey', then user should setup API_TOKEN / TEAM_URL / AUTHORIZATION before running content-cli. Do you agree ?

promeris: The env variable authentication should be working the same as before with the TEAM_URL and API_TOKEN, where you can provide an Application or User token in the API_TOKEN variable. However you would need somehow set the new prefix property for the profile [here](https://github.com/celonis/content-cli/blob/master/src/services/profile.service.ts#L72), because we create a profile from the env variables. 

To achieve this you can do this approach of testing out the token by making a request to `/api/cloud` and check if the prefix to be used should be `Bearer` or `AppKey`. I hope this clears your confusion.

kaztakata: @promeris 
I got your point. 
When env variables API_TOKEN / TEAM_URL are passed, then call `authenticate` method to build headers.
If env variables are not there, then use Profile file and get api_token, team_url and new attribute ('Bearer' or 'AppKey').

I will do update until next week.

src/validators/profile.validator.ts
```diff
@@ -35,10 +34,17 @@ export class ProfileValidator {
...
+                    request.get(url, options, (err, res) => {
+                        if (res.statusCode === 200) {
+                            resolve("AppKey");
```
kaztakata, line 40: return authenticationType when authentication is succeeded.

src/commands/profile.command.ts
```diff
@@ -12,7 +12,7 @@ export class ProfileCommand {
         profile.name = await QuestionService.ask("Name of the profile: ");
         profile.team = await QuestionService.ask("Your team (please provide the full url): ");
         profile.apiToken = await QuestionService.ask("Your api token: ");
-        await ProfileValidator.validateProfile(profile);
+        profile.authenticationType = await ProfileValidator.validateProfile(profile);
```
kaztakata, line 15: store autheticationType to profile.

src/services/http-client.service.ts
```diff
@@ -42,44 +42,75 @@ export class HttpClientService {
...
         return new Promise<any>((resolve, reject) => {
-            request.get(url, this.makeOptions(profile), (err, res) => {
+            request.get(url, this.makeOptions(headers, null), (err, res) => {
                 this.handleResponse(res, resolve, reject);
             });
         });
     }
 
-    private makeOptions(profile: Profile, body: object = {}): CoreOptions {
+    public async buildRequestHeadersWithAuthentication(profile: Profile): Promise<any> {
```
kaztakata, line 61: buildRequestHeadersWithAuthentication method is called before main method (pull/push etc.), first check if authenticationType exists. If exists, create headers and return it. Else check both 'Bearer' and 'AppKey' then return authenticated headers.

kaztakata:
@promeris 
I updated program. Could you review it ?

src/services/http-client.service.ts
promeris: I think we can simplify this part, so that we completely remove this logic of checking for authenticationType in the http-client, and just do it when creating the profile. So we have two cases when the authenticationType doesn't exists: 

1. Using content-cli with env variables
2. Profiles that were created before these changes don't have the authenticationType

For the first point I think it would be better to make the same change as when creating a new profile [here](https://github.com/celonis/content-cli/pull/71/files#diff-b85bcaaabf1adeb99ad93e463f998f0d61bc5e66b80052e11b38589929ba1d3fR70).
```typescript
private async buildProfileFromEnvVariables(): Promise<Profile> {
        const profileVariables = this.getProfileEnvVariables();
        const profile: Profile = {
            name: profileVariables.teamUrl,
            team: profileVariables.teamUrl,
            apiToken: profileVariables.apiToken,
            authenticationType: AuthenticationType.BEARER
        };
        profile.authenticationType = await ProfileValidator.validateProfile(profile);
        return profile;
    }
```

For the second point, you can use the fact that the previously created profiles have authenticationType `Bearer`, and substitute the `buildRequestHeadersWithAuthentication()` method with another like the following:
```typescript
private buildAuthorizationHeaders(profile: Profile): Headers {
        const authenticationType = profile.authenticationType || AuthenticationType.BEARER;
        return {
            authorization: `${authenticationType} ${apiToken}`,
            "content-type": "application/json",
        };
    }
```
That first statement in the method would make sure that if a profile doesn't have authenticationType defined, it would use the `Bearer` one by default, and solve the problem.

After these changes you wouldn't need to do this call after every command (push, pull), but just call the `buildAuthorizationHeaders` method to create the headers like before.

src/validators/profile.validator.ts
```diff
@@ -35,10 +34,17 @@ export class ProfileValidator {
...
+                        } else {
+                            logger.error(new FatalError("The provided team or api key is wrong."));
+                            reject();
+                        }
+                    });
                 } else {
-                    resolve();
+                    resolve("Bearer");
```
promeris, line 47: I also have one improvement instead of using `Bearer` and `AppKey` strings everywhere. You can make a type for the authenticationType that only allows `Bearer` or `AppKey`. You can insert the following snippet in the `profile.interface.ts` file.
```typescript
export interface Profile {
    name: string;
    team: string;
    apiToken: string;
    authenticationType: AuthenticationType
}

export type AuthenticationType = "Bearer" | "AppKey";
// tslint:disable-next-line:variable-name
export const AuthenticationType: { [key: string]: AuthenticationType } = {
    BEARER: "Bearer",
    APPKEY: "AppKey"
}
```
This would make sure that the value of the authenticationType is always within the defined values and you can also use these constants instead of the plain strings. For example here you would use `AuthenticationType.BEARER` instead of `"Bearer"`.

promeris (commented):
Hey @kaztakata, thanks for doing the suggested changes. There are a few things left to do and I left comments with instructions. After doing these last changes I think the feature is there and ready to be merged.

kaztakata:
@promeris 
I updated program according to your instruction. In the end I deleted most of code I added before and now it is simply united.

src/interfaces/profile.interface.ts
```diff
@@ -2,4 +2,12 @@ export interface Profile {
     name: string;
     team: string;
     apiToken: string;
+    authenticationType: string;
```
promeris, line 5: nit: You can use the AuthenticationType here as well, like `authnenticationType: AuthenticationType` :smile:

promeris (dismissed):
I tested it out myself and the authentication with AppKey is working correctly. I left one small comment that you can change if you want, but other than that the PR looks good. Feel free to merge the PR. Thanks for taking the time to contribute to the project.

src/interfaces/profile.interface.ts
kaztakata: OK I changed it.

kaztakata:
@promeris 
Thanks for you sincere review. Please check my last change, then I will merge my update.

promeris (approved):

kaztakata:
@promeris 
I do not have write access to this repository, so could you merge my change instead of me ?

PR #80, opened 2022-01-11
[FEAT] Allow pushing javascript sourcemaps to celonis DataDog

yaabed:
#### Description

Add WidgetSourcemapsCommand to allow pushing sourcemaps to celonis datadog.

#### Checklist

### DIFF
--- a/src/commands/widget-sourcemaps.command.ts
+++ b/src/commands/widget-sourcemaps.command.ts
@@ -3,9 +3,9 @@ import { WidgetSourcemapsManagerFactory } from "../content/factory/widget-source
 
 export class WidgetSourcemapsCommand {
     private contentService = new ContentService();
-    private widgetManagerFactory = new WidgetSourcemapsManagerFactory();
+    private widgetSourcemapsFactory = new WidgetSourcemapsManagerFactory();
 
     public async pushSourceMaps(profile: string): Promise<void> {
-        await this.contentService.push(profile, this.widgetManagerFactory.createManager());
+        await this.contentService.push(profile, this.widgetSourcemapsFactory.createManager());
     }
 }
--- a/src/commands/widget-sourcemaps.command.ts
+++ b/src/commands/widget-sourcemaps.command.ts
@@ -5,10 +5,7 @@ export class WidgetSourcemapsCommand {
     private contentService = new ContentService();
     private widgetManagerFactory = new WidgetSourcemapsManagerFactory();
 
-    public async pushWidget(profile: string, tenantIndependent: boolean, userSpecific: boolean): Promise<void> {
-        await this.contentService.push(
-            profile,
-            this.widgetManagerFactory.createManager(tenantIndependent, userSpecific)
-        );
+    public async pushSourceMaps(profile: string): Promise<void> {
+        await this.contentService.push(profile, this.widgetManagerFactory.createManager());
     }
 }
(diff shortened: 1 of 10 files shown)


src/commands/widget-sourcemaps.command.ts
```diff
@@ -0,0 +1,14 @@
+import { ContentService } from "../services/content.service";
+import { WidgetSourcemapsManagerFactory } from "../content/factory/widget-sourcemaps-manager.factory";
+
+export class WidgetSourcemapsCommand {
+    private contentService = new ContentService();
+    private widgetManagerFactory = new WidgetSourcemapsManagerFactory();
```
GencBlakqoriP9, line 6: This should be named `widgetSourceMapFactory`.

src/content/factory/widget-sourcemaps-manager.factory.ts
```diff
@@ -0,0 +1,24 @@
+import * as path from "path";
+import { FatalError, logger } from "../../util/logger";
+import { WidgetSourcemapsManager } from "../manager/widget-sourcemaps.manager";
+import { WidgetManagerFactory } from "./widget-manager.factory";
+
+export class WidgetSourcemapsManagerFactory {
+    private widgetManagerFactory = new WidgetManagerFactory();
+
+    public createManager(tenantIndependent: boolean = false, userSpecific: boolean = false): WidgetSourcemapsManager {
```
GencBlakqoriP9, line 9: `tenantIndependent` and `userSpecific` seem not to be used at all?

src/content-cli-push.ts
```diff
@@ -91,6 +93,12 @@ class Push {
             .option("--packageManager", "Upload widget to package manager (deprecated)") // Deprecated
             .action(async cmd => {
                 await new WidgetCommand().pushWidget(cmd.profile, !!cmd.tenantIndependent, !!cmd.userSpecific);
+                await new WidgetSourcemapsCommand().pushWidget(
```
GencBlakqoriP9, line 96: I think we should add a flag for this and not upload it always. And also, maybe we should make it dependent if `--tenantIndependent` is being used as well, because idk if it makes that much sense to push the source maps without that flag being true.

GencBlakqoriP9 (changes requested):

src/content-cli-push.ts
yaabed: we don't push if the environment variables are not set. Check widget-sourcemaps.manager.ts `pushWidgetSourcemaps` method.

src/content/factory/widget-sourcemaps-manager.factory.ts
yaabed: Yeah I will remove them

src/content-cli-push.ts
GencBlakqoriP9: Ah okay I haven't seen that. Personally I think it's nicer to have a separate flag rather than making it dependent on the environment variables, because it's not clear that whenever you use env variables => it pushes something to datadog. Additionally, that still doesn't ensure that the source maps are not pushed when we push only to a specific team or a user.

yaabed: Yeah as you wish, I don't have a good debate. I will add that as a flag.

GencBlakqoriP9 (dismissed):

src/commands/widget-sourcemaps.command.ts
yaabed: 60e772bfccc6eca503e06dc2e9dc0d10041f05d1

GencBlakqoriP9 (dismissed):

qmucolli (dismissed):

qmucolli:
Can we bump the minor version? It seems like a bigger change than just upgrading the patch

qmucolli (approved):

PR #98, opened 2022-07-12
bugfix TN make profile validation more reliable for application keys

promeris:
#### Description

Changed to using team endpoint (`api/cloud/team`) for validating the api token as when using application token the previous endpoint would return empty body for user token validation.

#### Checklist

### DIFF
--- a/src/validators/profile.validator.ts
+++ b/src/validators/profile.validator.ts
@@ -3,6 +3,10 @@ import { FatalError, logger } from "../util/logger";
 import validUrl = require("valid-url");
 import request = require("request");
 
+interface CloudResponse {
+    domain: string;
+}
+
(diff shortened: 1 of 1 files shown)


src/validators/profile.validator.ts
```diff
@@ -49,4 +48,11 @@ export class ProfileValidator {
             });
         });
     }
+
+    private static parseBody(responseBody: string): CloudResponse {
+        try {
+            return JSON.parse(responseBody);
+            // tslint:disable-next-line:no-empty
+        } catch (ignored) {}
```
meritonhusaj, line 56: does logger error help us here in case we need to debug this ever?

meritonhusaj (approved):

PR #119, opened 2023-05-29
TN-3954: Add content cli flag for inlcudingDependencies in package list

fisgeci:
#### Description

Added flag to allow downloading of package list with dependencies and variables
#### Checklist


src/interfaces/save-content-node.interface.ts
```diff
@@ -1,5 +1,9 @@
+import {AssetMetadataTransport} from "../api/package-manager-api";
+
 export interface SaveContentNode {
+    id: string,
```
promeris, line 4: I think this is not needed anymore as well, since we are using the new complete transports.

src/api/package-manager-api.ts
```diff
@@ -0,0 +1,76 @@
...
+    key: string;
+    name: string;
+    rootNodeKey: string;
+    assetType: string;
+    nodeType: string;
+    parentNodeId: string;
+    invalidContent: string;
+    serializedContent: string;
+    serializationType: string;
+    draftId: string; // TODO - Delete after the external services have adapted to the change
```
promeris, line 18: Let's keep here only the fields we are explicitly using in the CLI. Also the comment should not be needed.

src/services/package-manager/package-service.ts
```diff
@@ -0,0 +1,95 @@
...
+    protected readonly fileDownloadedMessage = "File downloaded successfully. New filename: ";
+
+    public async listPackages(): Promise<void> {
+        const nodes = await this.findAllPackages();
+        nodes.forEach(node => {
+            logger.info(`${node.name} - Key: "${node.key}"`);
+        });
+    }
+
+    public findAllPackages(): Promise<ContentNodeTransport[]> {
```
promeris, line 18: I think this is not used anywhere outside this class, but also can be directly used as `packageManagerApi.findAllPackages()` where needed.

src/services/package-manager/package-service.ts
```diff
@@ -0,0 +1,95 @@
...
+            dependenciesByPackageKey.set(packageWithDependency.rootNodeId, dependenciesOfPackage);
+        });
+
+        return dependenciesByPackageKey
+    }
+
+    public getPackagesWithDependencies(nodeIds: string[], draftIdsByNodeId: Map<string, string>): Promise<PackageDependencyTransport[]> {
+        return new Promise(async resolve => {
+            const results = await Promise.all(nodeIds.map(async nodeId => {
+                return packageManagerApi.findDependenciesOfPackage(nodeId, draftIdsByNodeId.get(nodeId));
```
promeris, line 76: I think we will need a new API that returns all this information with a single call, but we can work with this for now. Let's add it to the TODO list

promeris (commented):
Last point, please bump the patch version in package.json as the workflow is not working correctly currently. Other than that looks good.

src/api/package-manager-api.ts
```diff
@@ -0,0 +1,64 @@
...
+    variableAssignments: VariablesAssignments[]
+}
+
+export interface VariablesAssignments {
+    key: string;
+    value: object;
+    type: string;
+}
+
+class PackageManagerApi {
```
qmucolli, line 45: Should we maybe make this as PackageApi instead?

src/services/http-client-service.v2.ts
```diff
@@ -0,0 +1,70 @@
...
+    private buildAuthorizationHeaders(profile: Profile): Headers {
+        const authenticationType = profile.authenticationType || AuthenticationType.BEARER;
+        return {
+            authorization: `${authenticationType} ${profile.apiToken}`,
+            "content-type": "application/json",
+        };
+    }
+}
+
+export const httpClientV2 = new HttpClientServiceV2();
```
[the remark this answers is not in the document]
fisgeci: This is to ensure that the instance is singelton. So we can just import it where we need it and not create a new object each time.

src/services/package-manager/package-service.ts
```diff
@@ -0,0 +1,87 @@
...
+    protected readonly fileDownloadedMessage = "File downloaded successfully. New filename: ";
+
+    public async listPackages(): Promise<void> {
+        const nodes = await packageManagerApi.findAllPackages();
+        nodes.forEach(node => {
+            logger.info(`${node.name} - Key: "${node.key}"`);
+        });
+    }
+
+    public async findAndExportAllPackages(includeDependencies: boolean): Promise<void> {
```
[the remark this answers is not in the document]
fisgeci: same as here https://github.com/celonis/content-cli/pull/119#discussion_r1226370068. We will seperate the listing from the writing to file.

src/services/package-manager/package-service.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public async getPackagesWithDependencies(nodeIds: string[], draftIdsByNodeId: Map<string, string>): Promise<PackageDependencyTransport[]> {
+        const results = await Promise.all(nodeIds.map(async nodeId => {
+            return packageManagerApi.findDependenciesOfPackage(nodeId, draftIdsByNodeId.get(nodeId));
+        }))
+
+        const dependencies: PackageDependencyTransport[] = [];
+
+        results.forEach(packageDependencies => {
```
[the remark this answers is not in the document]
fisgeci: I was using .map  function which returned a map and had to extract the results. I removed that and returning the results directly now.

src/api/package-api.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    public async findAllPackagesWithVariableAssignments(): Promise<PackageWithVariableAssignments[]> {
+        return httpClientV2.get("/package-manager/api/packages/with-variable-assignments")
+    }
+
+    public async findLatestVersionById(nodeId: string): Promise<PackageHistoryTransport> {
+        return httpClientV2.get(`/package-manager/api/packages/${nodeId}/latest-version`);
+    }
+
+    public async findAssignedDatamodels(packageKey: string) : Promise<StudioDataModelTransport[]> {
```
promeris, line 30: This should be in a separate `ComputePoolsApi` class.

src/api/package-api.ts
```diff
@@ -0,0 +1,36 @@
...
+} from "../interfaces/package-manager.interfaces";
+
+
+
+class PackageApi {
+    public async findAllPackages(): Promise<ContentNodeTransport[]> {
+        return httpClientV2.get("/package-manager/api/packages");
+    }
+
+    public async findAllNodesOfType(assetType?: string): Promise<ContentNodeTransport[]> {
```
promeris, line 14: This should be in a separate `NodeApi` class.

src/api/package-api.ts
```diff
@@ -0,0 +1,36 @@
...
+        return httpClientV2.get(`/package-manager/api/nodes?assetType=${assetType}`)
+    }
+
+    public async findDependenciesOfPackage(nodeId: string, draftId: string): Promise<PackageDependencyTransport[]> {
```
promeris, line 18: This should be in a separate `PackageDependenciesApi` class.

src/content-cli-list.ts
```diff
@@ -1,18 +1,26 @@
...
             .description("Command to list all packages")
             .option("-p, --profile <profile>", "Profile which you want to use to list packages")
             .option("--json", "Return response as json type", "")
+            .option("--includeDependencies", "Include variables and dependencies", "")
             .action(async cmd => {
-                await new PackageCommand().listPackages(cmd.profile, cmd.json);
+                try {
+                    await new PackageCommand().listPackages(cmd.json, cmd.includeDependencies)
+                } catch (err) {
+                    logger.error(new FatalError(err));
```
promeris, line 22: I think we shouldn't throw plain errors to the users, but catch these where thrown (probably api calls) and display more some better information. We can also keep this AI for afterwards.

promeris (dismissed):

src/services/http-client-service.v2.ts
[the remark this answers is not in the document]
qmucolli: Yes, however is this the way that singleton is handled in node?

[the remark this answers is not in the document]
fisgeci: Yes.

src/api/compute-pool-api.ts
```diff
@@ -0,0 +1,16 @@
+import {StudioDataModelTransport} from "../interfaces/package-manager.interfaces";
+import {httpClientV2} from "../services/http-client-service.v2";
+import {FatalError} from "../util/logger";
+
+class ComputePoolApi {
+    public static readonly INSTANE = new ComputePoolApi();
```
qmucolli, line 6: Should this be instance?

src/api/package-api.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    public async findLatestVersionById(nodeId: string): Promise<PackageHistoryTransport> {
+        return httpClientV2.get(`/package-manager/api/packages/${nodeId}/latest-version`).catch(e => {
+            throw new FatalError(`Problem getting latest version of package: ${e}`);
+        });
+    }
+
+}
+
+export const packageApi = new PackageApi();
```
qmucolli, line 29: Should we do the INSTANCE pattern here as well?

qmucolli (dismissed):

qmucolli (approved):

PR #121, opened 2023-06-14
PAD-1915: new command introduced to listing assets

frkntrcnn:
#### Description

Please include a description of the change and why the change was made.

- New command introduced for the asset listings along with assetType option.
- Fetching with yml option has been added.

#### Checklist

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -288,7 +288,7 @@ assets you have access to. 
 -   It is also possible to download assets in YML format by adding '--yml' option.
 
 ```
-content-cli list packages -p <your-chosen-profile>
+content-cli list assets -p <your-chosen-profile> --assetType SCENARIO
 ```
 
 ### Asset options for Analysis
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -285,7 +285,7 @@ With this command you can retrieve a list of all assets with the --assetType o
 The command takes your permissions into consideration and only lists the
 assets you have access to. 
 
--   It is also possible to download assets in YML format by adding '--yml' option.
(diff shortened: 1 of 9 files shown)


DOCUMENTATION.md
```diff
@@ -279,6 +279,18 @@ packages you have access to. 
 content-cli list packages -p <your-chosen-profile>
 ```
 
+### List all assets in Studio 
+
+With this command you can retrieve a list of all assets with the --assetType option within a team.
+The command takes your permissions into consideration and only lists the
+assets you have access to. 
+
+-   It is also possible to download assets in YML format by adding '--yml' option.
```
promeris, line 288: Why are we doing YAML now, can we just keep it in JSON like the other responses?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@celonis/content-cli",
-  "version": "0.3.0",
+  "version": "0.3.1",
```
promeris, line 3: Can you pull latest `master` branch, because I think we are already at `0.3.2` version.

DOCUMENTATION.md
frkntrcnn: @promeris seems yml responses have been used for the assets. The writing file base function uses [this overwritten method](https://github.com/celonis/content-cli/blob/b4d34703173838f19d622405cd760ed63b5ad4dd/src/content/manager/asset.manager.ts#L123). Wanted to stick with it. Otherwise there will be a difference between pull & list assets cmds response types.

frkntrcnn:
@fisgeci @promeris as Fis suggested, I've migrated these changes to the new structure which you guys recently implemented. FYI ℹ️

fisgeci (approved):

promeris (approved):

PR #124, opened 2023-06-25
TN-3955: Introduce batch export

AlbinHoxha:
#### Description

We will introduce a new command to batch export all packages with option to also export package dependencies

#### Checklist

### DIFF
--- a/src/api/node-api.ts
+++ b/src/api/node-api.ts
@@ -11,13 +11,6 @@ class NodeApi {
                 throw new FatalError(`Problem getting nodes of type ${assetType}: ${e}`);
             });
     }
-
-    public async findAllByRootKey(rootKey: string): Promise<ContentNodeTransport[]> {
-        return httpClientV2.get(`/package-manager/api/nodes/by-root-key/${rootKey}`)
-            .catch(e=> {
-                throw new FatalError(`Problem getting nodes for root key ${rootKey}: ${e}`);
-            });
-    }
 }
 
 export const nodeApi = NodeApi.INSTANCE;
--- a/src/api/node-api.ts
+++ b/src/api/node-api.ts
@@ -12,7 +12,7 @@ class NodeApi {
             });
     }
 
-    public async findOneByKeyAndRootNodeKey(packageKey: string, nodeKey: string): Promise<ContentNodeTransport> {
+    public async findOneByKeyAndRootNodeKey(packageKey: string, nodeKey: string): Promise<ContentNodeTransport | null> {
         return httpClientV2.get(`/package-manager/api/nodes/${packageKey}/${nodeKey}`).catch(e => {
             return null
         });
--- a/src/api/node-api.ts
+++ b/src/api/node-api.ts
@@ -14,7 +14,7 @@ class NodeApi {
 
     public async findOneByKeyAndRootNodeKey(packageKey: string, nodeKey: string): Promise<ContentNodeTransport | null> {
         return httpClientV2.get(`/package-manager/api/nodes/${packageKey}/${nodeKey}`).catch(e => {
-            return null
+            return null;
         });
     }
 }
--- a/src/api/node-api.ts
+++ b/src/api/node-api.ts
@@ -7,9 +7,15 @@ class NodeApi {
 
     public async findAllNodesOfType(assetType?: string): Promise<ContentNodeTransport[]> {
         return httpClientV2.get(`/package-manager/api/nodes?assetType=${assetType}`)
-            .catch(e=> {
+            .catch(e => {
                 throw new FatalError(`Problem getting nodes of type ${assetType}: ${e}`);
             });
+    }
+
+    public async findOneByKeyAndRootNodeKey(packageKey: string, nodeKey: string): Promise<ContentNodeTransport> {
+        return httpClientV2.get(`/package-manager/api/nodes/${packageKey}/${nodeKey}`).catch(e => {
+            return null
+        });
     }
 }
 
--- a/src/api/node-api.ts
+++ b/src/api/node-api.ts
@@ -11,6 +11,13 @@ class NodeApi {
                 throw new FatalError(`Problem getting nodes of type ${assetType}: ${e}`);
             });
     }
+
+    public async findAllByRootKey(rootKey: string): Promise<ContentNodeTransport[]> {
+        return httpClientV2.get(`/package-manager/api/nodes/by-root-key/${rootKey}`)
+            .catch(e=> {
+                throw new FatalError(`Problem getting nodes for root key ${rootKey}: ${e}`);
+            });
+    }
 }
 
 export const nodeApi = NodeApi.INSTANCE;
--- a/src/api/package-api.ts
+++ b/src/api/package-api.ts
@@ -39,17 +39,14 @@ class PackageApi {
         });
     }
 
-    public async findOneByKeyAndRootNodeKey(packageKey: string): Promise<ContentNodeTransport> {
-        return httpClientV2.get(`/package-manager/api/nodes/${packageKey}/${packageKey}`).catch(e => {
-            return null
-        });
-    }
-
     public async importPackage(nodeContent: any, nodeId: string, spaceId: string, overwrite: boolean): Promise<any> {
-        await httpClientV2.post(`/package-manager/api/packages/import?spaceId=${spaceId}` + (overwrite ? "&overwrite=true" : ""), nodeContent).catch(e => {
+        await httpClientV2.postFile("/package-manager/api/packages/import", nodeContent, {
+            spaceId: spaceId,
+            overwrite: overwrite
+        }).catch(e => {
             throw new FatalError(`Problem importing package: ${e}`);
         });
-        if(overwrite) {
+        if (overwrite) {
             return await httpClientV2.put(`/package-manager/api/packages/${nodeId}/move/${spaceId}`, {}).catch(e => {
                 throw new FatalError(`Problem moving package: ${e}`);
             });
@@ -63,7 +60,7 @@ class PackageApi {
     }
 
     public async publishPackage(activatePackage: ActivatePackageTransport): Promise<void> {
-        await httpClientV2.postJson(`/package-manager/api/packages/${activatePackage.packageKey}/activate`, activatePackage).catch(e => {
+        await httpClientV2.post(`/package-manager/api/packages/${activatePackage.packageKey}/activate`, activatePackage).catch(e => {
             throw new FatalError(`Problem activating package with key ${activatePackage.packageKey}: ${e}`);
         });
     }
--- a/src/api/package-api.ts
+++ b/src/api/package-api.ts
@@ -39,23 +39,18 @@ class PackageApi {
         });
     }
 
-    public async importPackage(nodeContent: any, nodeId: string, spaceId: string, overwrite: boolean): Promise<any> {
+    public async importPackage(nodeContent: any, spaceId: string, overwrite: boolean): Promise<any> {
         await httpClientV2.postFile("/package-manager/api/packages/import", nodeContent, {
             spaceId: spaceId,
             overwrite: overwrite
         }).catch(e => {
             throw new FatalError(`Problem importing package: ${e}`);
         });
-        if (overwrite) {
-            return await httpClientV2.put(`/package-manager/api/packages/${nodeId}/move/${spaceId}`, {}).catch(e => {
-                throw new FatalError(`Problem moving package: ${e}`);
-            });
-        }
     }
 
-    public async updatePackageDependency(nodeId: string, packageDependency: PackageDependencyTransport): Promise<void> {
-        await httpClientV2.put(`/package-manager/api/package-dependencies/${nodeId}/dependency/by-key/${packageDependency.key}`, packageDependency).catch(e => {
-            throw new FatalError(`Problem updating package dependency: ${e}`);
+    public async movePackageToSpace(nodeId: string, spaceId: string): Promise<void> {
+        await httpClientV2.put(`/package-manager/api/packages/${nodeId}/move/${spaceId}`, {}).catch(e => {
+            throw new FatalError(`Problem moving package: ${e}`);
         });
     }
 
--- a/src/api/package-api.ts
+++ b/src/api/package-api.ts
@@ -15,6 +15,10 @@ class PackageApi {
         });
     }
 
+    public async exportPackage(rootPackageKey: string): Promise<Buffer> {
+        return await httpClientV2.downloadFile(`/package-manager/api/packages/${rootPackageKey}/export?newKey=${rootPackageKey}`);
+    }
+
     public async findAllPackagesWithVariableAssignments(): Promise<PackageWithVariableAssignments[]> {
         return httpClientV2.get("/package-manager/api/packages/with-variable-assignments").catch(e => {
             throw new FatalError(`Problem getting variables of packages: : ${e}`);
--- a/src/api/package-api.ts
+++ b/src/api/package-api.ts
@@ -37,6 +37,12 @@ class PackageApi {
         });
     }
 
+    public async findActiveVersionByIds(nodeIds: string[]): Promise<PackageHistoryTransport[]> {
+        return httpClientV2.post("/package-manager/api/packages/active/by-ids", nodeIds).catch(e => {
+            throw new FatalError(`Problem getting latest version of packages: ${e}`);
+        });
+    }
+
     public async findNextVersion(nodeId: string): Promise<PackageHistoryTransport> {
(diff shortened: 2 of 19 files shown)


src/api/node-api.ts
```diff
@@ -11,6 +11,13 @@ class NodeApi {
                 throw new FatalError(`Problem getting nodes of type ${assetType}: ${e}`);
             });
     }
+
+    public async findAllByRootKey(rootKey: string): Promise<ContentNodeTransport[]> {
```
promeris, line 15: I don't see any usage of this. Can you check if we can delete it?

src/api/space-api.ts
```diff
@@ -0,0 +1,21 @@
...
+    PackageWithVariableAssignments
+} from "../interfaces/package-manager.interfaces";
+import {httpClientV2} from "../services/http-client-service.v2";
+import {FatalError} from "../util/logger";
+import {SpaceExportTransport} from "../interfaces/save-space.interface";
+
+class SpaceApi {
+    public static readonly INSTANCE = new SpaceApi();
+
+    public async findParentSpace(spaceId: string): Promise<SpaceExportTransport> {
```
promeris, line 13:
```suggestion
    public async findOne(spaceId: string): Promise<SpaceTransport> {
```

src/content-cli-export.ts
```diff
@@ -0,0 +1,44 @@
...
+import * as commander from "commander";
+
+export class Export {
+
+    public static packages(program: CommanderStatic): CommanderStatic {
+        program
+            .command("packages")
+            .description("Command to export all given packages")
+            .option("-p, --profile <profile>", "Profile which you want to use to list packages")
+            .option("--packageKeys <packageKeys...>", "Exports only given package keys", "[]")
```
promeris, line 15:
```suggestion
            .requiredOption("--packageKeys <packageKeys...>", "Exports only given package keys", "[]")
```

src/services/package-manager/package-service.ts
```diff
@@ -52,9 +63,25 @@ class PackageService {
...
+    public async batchExportPackages(packageKeys: string[], includeDependencies: boolean, profileName: string): Promise<void> {
+        const allPackages = await packageApi.findAllPackages();
+        let nodesListToExport: BatchExportNodeTransport[] = allPackages.filter(node => packageKeys.includes(node.key));
+        const actionFlowsPackageKeys = (await nodeApi.findAllNodesOfType("SCENARIO")).map(node => node.rootNodeKey);
+        nodesListToExport = nodesListToExport.filter(node => !actionFlowsPackageKeys.includes(node.key));
+
+        if (includeDependencies) {
+            nodesListToExport = await this.fillNodeDependencies(nodesListToExport, allPackages, actionFlowsPackageKeys);
+            nodesListToExport = await spaceService.getParentSpaces(nodesListToExport);
+            await this.exportToZip(nodesListToExport, profileName);
```
promeris, line 97: Let's move this after the if block.

src/services/package-manager/package-service.ts
```diff
@@ -52,9 +63,25 @@ class PackageService {
...
+        } else {
+            nodesListToExport = await spaceService.getParentSpaces(nodesListToExport);
+            await this.exportToZip(nodesListToExport, profileName);
+        }
+    }
+
+    public async getPackagesDependenciesByPackageId(nodeIds: string[], draftIdByNodeId: Map<string, string>, actionFlowPackageKeys: string[]): Promise<Map<string, PackageDependencyTransport[]>> {
```
promeris, line 104: Let's make this private as it's only used in this class.

src/services/package-manager/package-service.ts
```diff
@@ -78,23 +105,126 @@ class PackageService {
...
+            listOfDependencies = listOfDependencies.filter(dependency => actionFlowPackageKeys.includes(dependency.key));
+        });
         const dependencies: PackageDependencyTransport[] = [];
         results.forEach(result => dependencies.push(...result));
 
         return dependencies;
     }
 
+    private async fillNodeDependencies(nodesListToExport: BatchExportNodeTransport[], allPackages: ContentNodeTransport[], actionFlowPackageKeys: string[]): Promise<BatchExportNodeTransport[]> {
+        nodesListToExport = await this.getVersionOfPackages(nodesListToExport);
```
promeris: Let's not mutate the given method parameters and create a new reference.

src/services/package-manager/space-service.ts
```diff
@@ -0,0 +1,26 @@
+import {SpaceExportTransport} from "../../interfaces/save-space.interface";
+import {spaceApi} from "../../api/space-api";
+import {ContentNodeTransport} from "../../interfaces/package-manager.interfaces";
+import {promises} from "fs";
```
promeris, line 4: nit: unused import

src/services/package-manager/package-service.ts
```diff
@@ -78,23 +105,126 @@ class PackageService {
...
         fileService.writeToFileWithGivenName(JSON.stringify(nodes, fieldsToInclude), filename);
         logger.info(FileService.fileDownloadedMessage + filename);
     }
+
+    private async exportPackagesAndAssets(nodes: BatchExportNodeTransport[], profileName: string): Promise<any[]> {
+        const zips = [];
+        const packages = nodes as ContentNodeTransport[];
+        for (const rootPackage of packages) {
+            const profile = await this.profileService.findProfile(profileName)
+            const exportedPackage = await this.httpClientService.pullFileData(`${profile.team}/package-manager/api/packages/${rootPackage.key}/export`, profile);
```
promeris: Let's strictly use the v2 http client.

src/api/package-api.ts
```diff
@@ -26,6 +27,46 @@ class PackageApi {
...
+        });
+    }
+
+    public async findNextVersion(nodeId: string): Promise<PackageHistoryTransport> {
+        return httpClientV2.get(`/package-manager/api/packages/${nodeId}/next-version`).catch(e => {
+            throw new FatalError(`Problem getting latest version of package: ${e}`);
+        });
+    }
+
+    public async findOneByKeyAndRootNodeKey(packageKey: string): Promise<ContentNodeTransport> {
```
promeris, line 42: This belongs in NodeApi instead.

src/api/package-api.ts
```diff
@@ -26,6 +27,46 @@ class PackageApi {
...
+        return httpClientV2.get(`/package-manager/api/nodes/${packageKey}/${packageKey}`).catch(e => {
+            return null
+        });
+    }
+
+    public async importPackage(nodeContent: any, nodeId: string, spaceId: string, overwrite: boolean): Promise<any> {
+        await httpClientV2.post(`/package-manager/api/packages/import?spaceId=${spaceId}` + (overwrite ? "&overwrite=true" : ""), nodeContent).catch(e => {
```
promeris, line 49: Instead of building URI strings, can we pass in the request params to the httpClient and let it handle building the URIs.

src/api/space-api.ts
```diff
@@ -0,0 +1,33 @@
...
+    PackageWithVariableAssignments
+} from "../interfaces/package-manager.interfaces";
+import {httpClientV2} from "../services/http-client-service.v2";
+import {FatalError} from "../util/logger";
+import {SaveSpace, SpaceExportTransport} from "../interfaces/save-space.interface";
+
+class SpaceApi {
+    public static readonly INSTANCE = new SpaceApi();
+
+    public async findOne(spaceId: string): Promise<SpaceExportTransport> {
```
promeris, line 13: Can we use a single `SpaceTransport` for all these methods?

src/content-cli-import.ts
```diff
@@ -0,0 +1,44 @@
...
+import * as commander from "commander";
+
+export class Import {
+
+    public static packages(program: CommanderStatic): CommanderStatic {
+        program
+            .command("packages")
+            .description("Command to import all given packages")
+            .option("-p, --profile <profile>", "Profile which you want to use to list packages")
+            .option("--spaceMapping <spaceMapping...>", "Mapping on which space packages will be created", "[]")
```
promeris, line 14:
```suggestion
            .option("--spaceMappings <spaceMappings...>", "List of mappings for importing packages to different target spaces. Mappings should follow format 'packageKey:targetSpaceKey'", "[]")
```

src/content-cli-import.ts
```diff
@@ -0,0 +1,44 @@
...
+            .option("--exportedDatapoolsFile <exportedDatapoolsFile>", "Exported datapool file (relative path)", "")
```
promeris, line 15: Let's discuss this a bit more, as the data pools should probably be included in the same export file. @frkntrcnn FYI

src/interfaces/package-manager.interfaces.ts
```diff
@@ -26,6 +32,10 @@ export interface PackageDependencyTransport {
     name: string;
     version: string;
     rootNodeId: string;
+    external: boolean
+    draftId: string;
+    updateAvailable: boolean;
```
promeris, line 37: Can we include this field in the listing as well? @frkntrcnn this can be used directly to warn users of dependencies not using latest active version.

src/api/package-api.ts
```diff
@@ -26,6 +27,46 @@ class PackageApi {
...
+    }
+
+    public async updatePackageDependency(nodeId: string, packageDependency: PackageDependencyTransport): Promise<void> {
+        await httpClientV2.put(`/package-manager/api/package-dependencies/${nodeId}/dependency/by-key/${packageDependency.key}`, packageDependency).catch(e => {
+            throw new FatalError(`Problem updating package dependency: ${e}`);
+        });
+    }
+
+    public async publishPackage(activatePackage: ActivatePackageTransport): Promise<void> {
+        await httpClientV2.postJson(`/package-manager/api/packages/${activatePackage.packageKey}/activate`, activatePackage).catch(e => {
```
promeris, line 66: This should be using normal `httpClientV2.post`, and for pushing files a separate method should be created like `postFile`.

src/services/http-client-service.v2.ts
```diff
@@ -15,10 +15,78 @@ class HttpClientServiceV2 {
...
+        return new Promise<any>((resolve, reject) => {
+            request.put(this.resolveUrl(url), this.makeOptionsJson(contextService.getContext().profile, JSON.stringify(body), "application/json;charset=utf-8"), (err, res) => {
+                this.handleResponse(res, resolve, reject);
+            });
+        }).catch(e => {
+            throw new FatalError(e);
+        });
+    }
+
+    public async pullFileData(url: string, profile: Profile): Promise<any> {
```
promeris, line 48: nit: maybe a better name is `downloadFile`?

src/services/package-manager/package-service.ts
```diff
@@ -52,9 +68,43 @@ class PackageService {
         this.exportListOfPackages(nodesListToExport, fieldsToInclude);
     }
 
-    public async getPackagesDependenciesByPackageId(nodeIds: string[], draftIdByNodeId: Map<string, string>): Promise<Map<string, PackageDependencyTransport[]>> {
+    public async batchImportPackages(spaceMapping: string[], exportedDatapoolsFile: string, exportedPackagesFile: string): Promise<void> {
+        if (exportedPackagesFile) {
```
promeris, line 72: This should be a required option, and we should remove the conditional block.

src/services/package-manager/package-service.ts
```diff
@@ -78,23 +112,235 @@ class PackageService {
...
         fileService.writeToFileWithGivenName(JSON.stringify(nodes, fieldsToInclude), filename);
         logger.info(FileService.fileDownloadedMessage + filename);
     }
+
+    private async exportPackagesAndAssets(nodes: BatchExportNodeTransport[], profileName: string): Promise<any[]> {
+        const zips = [];
+        const packages = nodes as ContentNodeTransport[];
+        for (const rootPackage of packages) {
+            const profile = await this.profileService.findProfile(profileName)
+            const exportedPackage = await httpClientV2.downloadFile(`${profile.team}/package-manager/api/packages/${rootPackage.key}/export?newKey=${rootPackage.key}`, profile);
```
fisgeci, line 290: Can we extract this into a api file.

src/services/package-manager/package-service.ts
```diff
@@ -78,23 +112,235 @@ class PackageService {
...
+            zips.push({
+                data: exportedPackage,
+                packageKey: rootPackage.key
+            });
+        }
+        return zips;
+    }
+
+    private async exportToZip(nodes: BatchExportNodeTransport[], profileName: string): Promise<void> {
```
fisgeci, line 299: I think we can remove profileName.

fisgeci (dismissed):

fisgeci (dismissed):

src/services/package-manager/package-service.ts
```diff
@@ -78,23 +112,242 @@ class PackageService {
...
+            node.datamodels = dataModelAssignments.get(node.key);
+        })
+
+        nodesListToExport = await this.getNodeDependencies(nodesListToExport, allPackages, actionFlowPackageKeys, nodePath);
+        return nodesListToExport
+    }
+
+    private async getNodeDependencies(nodesListToExport: BatchExportNodeTransport[], allPackages: ContentNodeTransport[], actionFlowPackageKeys: string[], nodePath: string[]): Promise<BatchExportNodeTransport[]> {
+        for (const node of nodesListToExport) {
+            const nodesToGetKeys = node.dependencies.filter(dependency => !nodesListToExport
```
fisgeci, line 266: Nit: maybe unexportedNodeKeys would fit this better. This name doesn't seem to fit in my opinion. What do you think?

src/interfaces/package-manager.interfaces.ts
frkntrcnn: @promeris agree, could be good to show it with just using this, instead of comparing versions

src/api/node-api.ts
```diff
@@ -7,10 +7,16 @@ class NodeApi {
...
         return httpClientV2.get(`/package-manager/api/nodes?assetType=${assetType}`)
-            .catch(e=> {
+            .catch(e => {
                 throw new FatalError(`Problem getting nodes of type ${assetType}: ${e}`);
             });
     }
+
+    public async findOneByKeyAndRootNodeKey(packageKey: string, nodeKey: string): Promise<ContentNodeTransport | null> {
+        return httpClientV2.get(`/package-manager/api/nodes/${packageKey}/${nodeKey}`).catch(e => {
+            return null
```
fisgeci, line 17: nit: add ;

fisgeci, line 17: Why do we return null here?

AlbinHoxha: Because there are cases when we cant find a node and it throws a 404 but for us that is a null value

src/api/package-api.ts
```diff
@@ -14,6 +15,10 @@ class PackageApi {
         });
     }
 
+    public async exportPackage(rootPackageKey: string): Promise<Buffer> {
+        return await httpClientV2.downloadFile(`/package-manager/api/packages/${rootPackageKey}/export?newKey=${rootPackageKey}`);
```
fisgeci, line 19: Can we add a catch clause to catch errors and throw a custom error here?

src/api/package-dependencies-api.ts
```diff
@@ -1,6 +1,6 @@
 import {PackageDependencyTransport} from "../interfaces/package-manager.interfaces";
 import {httpClientV2} from "../services/http-client-service.v2";
-import {FatalError} from "../util/logger";
+import {FatalError, logger} from "../util/logger";
```
fisgeci, line 3: nit: I think logger is not being used and can be removed.

src/services/package-manager/package-service.ts
```diff
@@ -2,13 +2,29 @@ import {logger} from "../../util/logger";
...
+import AdmZip = require("adm-zip");
+import {spaceApi} from "../../api/space-api";
+import * as path from "path";
+import {tmpdir} from "os";
+import {httpClientV2} from "../http-client-service.v2";
+import {SpaceTransport} from "../../interfaces/save-space.interface";
+import {ManifestDependency, ManifestNodeTransport} from "../../interfaces/manifest-transport";
 
 class PackageService {
+    private profileService = new ProfileService();
```
fisgeci, line 27: nit: this is  not needed i belive

src/api/node-api.ts
fisgeci: In this case do we want to continue? As i see from the code we are trying to access an id of a null object in this case. Will this throw an error?

fisgeci (approved):

AlbinHoxha:
We will be addressing the comments in another pr

fisgeci:
> We will be addressing the comments in another pr

https://github.com/celonis/content-cli/pull/126

PR #126, opened 2023-07-10
TN-4316: Export versioned packages.

fisgeci:
#### Description

When exporting packages and their dependencies, currently we only export the latest version of each package instead of the required version that some package along the chain depends on. 
This pr makes it so we export the used version of each package and also sets the field in the usedVersion field in the manifest for each package.

#### Checklist

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -261,7 +261,7 @@ You can use the export packages command to export multiple packages at once from
 
 ```
 //Exporting multiple packages at once
-content-cli export packages -p <your-chosen-profile> --packageKeys <packageKeysYouWantToExport>
+content-cli export packages -p <your-chosen-profile> --packageKeys <package1> <package2>
 ```
 
 Example usage would be:
@@ -270,6 +270,11 @@ content-cli export packages -p dev --packageKeys package-1 package-2
 ```
 
 You can use the --includeDependencies flag to also export the dependencies of the specified packages.
+
+```
+//Exporting multiple packages at once with dependencies
+content-cli export packages -p <your-chosen-profile> --packageKeys <package1> <package2> --includeDependencies
+```
 
 ### List all spaces in Studio
 With this command you can retrieve a list of all spaces within a team.
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -255,6 +255,22 @@ When you use overwrite the following is to be taken into consideration:
 content-cli push package -p my-profile-name --spaceKey my-space -f <path-to-my-local-package> --overwrite
 ```
 
+### Export multiple packages at once from Studio
+
+You can use the export packages command to export multiple packages at once from studio.
+
+```
+//Exporting multiple packages at once
+content-cli export packages -p <your-chosen-profile> --packageKeys <packageKeysYouWantToExport>
+```
+
+Example usage would be:
+```
+content-cli export packages -p dev --packageKeys package-1 package-2
+```
+
+You can use the --includeDependencies flag to also export the dependencies of the specified packages.
+
 ### List all spaces in Studio
 With this command you can retrieve a list of all spaces within a team.
 The command takes your permissions into consideration and only lists the
--- a/src/api/package-api.ts
+++ b/src/api/package-api.ts
@@ -1,7 +1,7 @@
 import {httpClientV2} from "../services/http-client-service.v2";
 import {
     ActivatePackageTransport,
-    ContentNodeTransport, PackageDependencyTransport, PackageHistoryTransport, PackageWithVariableAssignments
+    ContentNodeTransport, PackageHistoryTransport, PackageWithVariableAssignments
 } from "../interfaces/package-manager.interfaces";
 import {FatalError} from "../util/logger";
 
@@ -16,7 +16,11 @@ class PackageApi {
     }
 
     public async exportPackage(rootPackageKey: string, version?: string): Promise<Buffer> {
-        return await httpClientV2.downloadFile(`/package-manager/api/packages/${rootPackageKey}/export?newKey=${rootPackageKey}${version ? `&version=${version}` : ""}`);
+        const queryParams = new URLSearchParams();
+        queryParams.set("newKey", rootPackageKey);
+        queryParams.set("version", version ?? "");
+
+        return await httpClientV2.downloadFile(`/package-manager/api/packages/${rootPackageKey}/export?${queryParams.toString()}`);
     }
 
     public async findAllPackagesWithVariableAssignments(): Promise<PackageWithVariableAssignments[]> {
--- a/src/api/package-api.ts
+++ b/src/api/package-api.ts
@@ -20,7 +20,9 @@ class PackageApi {
         queryParams.set("newKey", rootPackageKey);
         queryParams.set("version", version ?? "");
 
-        return await httpClientV2.downloadFile(`/package-manager/api/packages/${rootPackageKey}/export?${queryParams.toString()}`);
+        return await httpClientV2.downloadFile(`/package-manager/api/packages/${rootPackageKey}/export?${queryParams.toString()}`).catch(e => {
+            throw new FatalError(`Pacakge ${rootPackageKey}_${version} failed to export.`)
+        });
     }
 
     public async findAllPackagesWithVariableAssignments(): Promise<PackageWithVariableAssignments[]> {
--- a/src/api/package-api.ts
+++ b/src/api/package-api.ts
@@ -15,8 +15,8 @@ class PackageApi {
         });
     }
 
-    public async exportPackage(rootPackageKey: string): Promise<Buffer> {
-        return await httpClientV2.downloadFile(`/package-manager/api/packages/${rootPackageKey}/export?newKey=${rootPackageKey}`);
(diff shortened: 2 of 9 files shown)


src/api/package-api.ts
```diff
@@ -15,8 +15,8 @@ class PackageApi {
         });
     }
 
-    public async exportPackage(rootPackageKey: string): Promise<Buffer> {
-        return await httpClientV2.downloadFile(`/package-manager/api/packages/${rootPackageKey}/export?newKey=${rootPackageKey}`);
+    public async exportPackage(rootPackageKey: string, version?: string): Promise<Buffer> {
+        return await httpClientV2.downloadFile(`/package-manager/api/packages/${rootPackageKey}/export?newKey=${rootPackageKey}${version ? `&version=${version}` : ""}`);
```
AlbinHoxha, line 19: can you look to use `querystring.stringify`to build parameters like we did in some other methods

src/commands/package.command.ts
```diff
@@ -48,7 +48,7 @@ export class PackageCommand {
         await packageService.batchExportPackages(packageKeys, includeDependencies);
     }
 
-    public async batchImportPackages(spaceMappings: string[], exportedDatapoolsFile: string, exportedPackagesFile: string): Promise<void> {
-        await packageService.batchImportPackages(spaceMappings ?? [], exportedDatapoolsFile, exportedPackagesFile);
+    public async batchImportPackages(spaceMappings: string[], dataModelMapping: string, exportedPackagesFile: string): Promise<void> {
```
AlbinHoxha, line 51: I think this should be a list of strings

src/content-cli-import.ts
```diff
@@ -12,10 +12,10 @@ export class Import {
             .description("Command to import all given packages")
             .option("-p, --profile <profile>", "Profile which you want to use to list packages")
             .option("--spaceMappings <spaceMappings...>", "List of mappings for importing packages to different target spaces. Mappings should follow format 'packageKey:targetSpaceKey'")
-            .option("--exportedDatapoolsFile <exportedDatapoolsFile>", "Exported datapool file (relative path)", "")
+            .option("--dataModelMapping <dataModelMapping>", "Data model mappings file (relative path)", "")
```
AlbinHoxha, line 15: did we not decide we will be sending a mapping like the spaceMapping and not the file?

src/services/package-manager/package-service.ts
```diff
@@ -264,40 +263,43 @@ class PackageService {
...
+
+        const fileName = "export_" + uuidv4() + ".zip";
+        zip.writeZip(fileName);
+        logger.info("Successfully exported package to: " + fileName);
     }
 
-    private exportManifestOfPackages(nodes: BatchExportNodeTransport[]): ManifestNodeTransport[] {
-        const manifestNodes: ManifestNodeTransport[] = [];
+    private exportManifestOfPackages(nodes: BatchExportNodeTransport[], dependencyVersionsByNodeKey: Map<string, string[]>): ManifestNodeTransport[] {
+        const manifestNodesByPackageKey = new Map<string, ManifestNodeTransport>();
```
AlbinHoxha, line 297: why is it better to have this as a map compared to a list?

src/services/package-manager/package-service.ts
```diff
@@ -264,40 +263,43 @@ class PackageService {
...
     }
 
-    private async exportPackagesAndAssets(nodes: BatchExportNodeTransport[]): Promise<any[]> {
+    private async exportPackagesAndAssets(nodes: BatchExportNodeTransport[], versionsByNodeKey: Map<string, string[]>): Promise<any[]> {
         const zips = [];
-        const packages = nodes as ContentNodeTransport[];
-        for (const rootPackage of packages) {
-            const exportedPackage = await packageApi.exportPackage(rootPackage.key)
+        for (const rootPackage of nodes) {
+            let exportedPackage;
```
AlbinHoxha, line 269: why do we need to declare it here instead of one line below

src/services/package-manager/package-service.ts
```diff
@@ -141,121 +140,121 @@ class PackageService {
...
+                    dependencyVersions.push(dependency.version);
+                    versionsByNodeKey.set(dependency.key, dependencyVersions);
+                }
+            });
 
-        nodesListToExport = await this.getNodeDependencies(nodesListToExport, allPackages, actionFlowPackageKeys, nodePath);
-        return nodesListToExport
-    }
+            const nodesToGetKeys = node.dependencies.filter(dependency => {
+                return !this.nodeHasBeenResolvedBefore(resolvedDependencies, dependency.key, dependency.version)
```
AlbinHoxha, line 233: Did we decide that we are safe from circular dependency and do not need to add a check that kills the process

src/services/package-manager/package-service.ts
```diff
@@ -141,121 +140,121 @@ class PackageService {
...
-        importedKeys.push(packageToImport.packageKey);
-        const nodeInTargetTeam = await nodeApi.findOneByKeyAndRootNodeKey(packageToImport.packageKey, packageToImport.packageKey);
-        await packageApi.importPackage(packageZip, targetSpaceId, !!nodeInTargetTeam);
-        if (nodeInTargetTeam) {
-            await packageApi.movePackageToSpace(nodeInTargetTeam.id, targetSpaceId)
-        }
-        await this.updateDependencyVersions(packageToImport);
-        await this.publishPackage(packageToImport);
-        logger.info(`Imported package with key: ${packageToImport.packageKey} successfully`)
+      // TODO [TN-4317](https://celonis.atlassian.net/browse/TN-4317)
```
AlbinHoxha, line 143: This pr is not related to import hence we do not need to comment everything out

src/content-cli-import.ts
fisgeci: We will recieve the mapping in the file. As opposed to proving the exported datapools as before.

src/services/package-manager/package-service.ts
fisgeci: It is faster to search by key when we need to update the version of each package.

src/services/package-manager/package-service.ts
fisgeci: I can just delete the import implementation as this pr breaks the current implementation. Ill just leave the comment that links the ticket.

src/services/package-manager/package-service.ts
fisgeci: This was left over.

src/services/package-manager/package-service.ts
```diff
@@ -266,38 +203,40 @@ class PackageService {
...
         for (const packageZip of packageZips) {
-            zip.addFile(packageZip.packageKey + ".zip", packageZip.data)
+            zip.addFile(`${packageZip.packageKey}_${packageZip.version}.zip`, packageZip.data)
         }
-        zip.writeZip("export_" + uuidv4() + ".zip");
-        logger.info("Successfully exported package");
+
+        const fileName = "export_" + uuidv4() + ".zip";
+        zip.writeZip(fileName);
+        logger.info("Successfully exported package to: " + fileName);
```
frkntrcnn, line 230: To stick with the structure, how would it be to use "File downloaded successfully. New filename: " this message on output?

src/interfaces/manifest-transport.ts
```diff
@@ -1,12 +1,12 @@
...
 
 export interface ManifestNodeTransport {
+    packageKeyAndVersion: string;
     packageKey: string;
     packageId: string,
-    packageVersion: string;
     variables: ManifestVariable[],
     space: ManifestSpace,
-    dependencies: ManifestDependency[]
+    dependenciesByVersion: Map<string, ManifestDependency[]>;
```
frkntrcnn, line 9: is there any possibility to pass dependency without specifying the version? Or should it be pointed out even if the latest version will be used?

src/commands/package.command.ts
promeris: There should probably be two options, either providing the output file from the import of dataModels (which can be implemented when that is ready), or you can define the mappings yourself like `--variableMappings dm-variable,dm-1;dm-variable2,dm-2`, etc.

DOCUMENTATION.md
```diff
@@ -255,6 +255,22 @@ When you use overwrite the following is to be taken into consideration:
 content-cli push package -p my-profile-name --spaceKey my-space -f <path-to-my-local-package> --overwrite
 ```
 
+### Export multiple packages at once from Studio
+
+You can use the export packages command to export multiple packages at once from studio.
+
+```
+//Exporting multiple packages at once
+content-cli export packages -p <your-chosen-profile> --packageKeys <packageKeysYouWantToExport>
```
promeris, line 264: Something like this maybe?
```suggestion
content-cli export packages -p <your-chosen-profile> --packageKeys <package1> <package2>
```

DOCUMENTATION.md
```diff
@@ -255,6 +255,22 @@ When you use overwrite the following is to be taken into consideration:
...
+```
+
+Example usage would be:
+```
+content-cli export packages -p dev --packageKeys package-1 package-2
+```
+
+You can use the --includeDependencies flag to also export the dependencies of the specified packages.
```
promeris, line 272: Let's add an example command for this as well.

promeris (approved):

PR #139, opened 2023-08-08
PAD-2091: Datapools success message refactors

frkntrcnn:
#### Description

Success message refactors for export & import operations of data pools to stick with the current output structure of content-cli

#### Checklist

### DIFF
--- a/src/services/data-pool/data-pool-service.ts
+++ b/src/services/data-pool/data-pool-service.ts
@@ -13,7 +13,7 @@ class DataPoolService {
         if (outputToJsonFile) {
             const reportFileName = "batch_import_report_" + uuidv4() + ".json";
             fileService.writeToFileWithGivenName(importReportString, reportFileName);
-            logger.info(FileService.fileDownloadedMessage + reportFileName);
+            logger.info("Batch import report file: " + reportFileName);
         } else {
             logger.info("Batch import report: \n" + importReportString);
         }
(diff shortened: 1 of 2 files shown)


src/services/data-pool/data-pool-service.ts
```diff
@@ -10,11 +10,10 @@ class DataPoolService {
         const importReport = await dataPoolApi.executeDataPoolsBatchImport(requestFileContent);
         const importReportString = JSON.stringify(importReport, null, 4);
 
-        logger.info("Data Pools batch import succeeded!");
         if (outputToJsonFile) {
             const reportFileName = "batch_import_report_" + uuidv4() + ".json";
             fileService.writeToFileWithGivenName(importReportString, reportFileName);
-            logger.info("Batch import report file: " + reportFileName);
+            logger.info(FileService.fileDownloadedMessage + reportFileName);
```
IvanGandacov, line 16: I honestly don't think the current message in this variable is too intuitive. 
`"File downloaded successfully. New filename: "`
The natural question is: Why was file downloaded if I just did an import? 
Also it is too generic.

I think having "Batch import report file" is more clear and tailored for this particular context.

frkntrcnn: true I did revert it

IvanGandacov (approved):

fisgeci (approved):

PR #141, opened 2023-08-09
Tn-4460: Add dataModelMappings option to import packages command

fisgeci:
#### Description

Added dataModelMappings assignment to package variables when importing packages and their dependencies.

#### Checklist

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -9,6 +9,8 @@
         -   [Pull package for EMS Store](#pull-package-for-ems-store)
     -   [Pull/Push individual assets from/to Studio](#pullpush-individual-assets-fromto-studio)
     -   [Overwrite Package In Studio](#overwrite-package-in-studio)
+    -   [Export multiple packages at once from Studio](#export-multiple-packages-at-once-from-studio)
+    -   [Importing multiple packages into Studio](#importing-multiple-packages-into-studio)
     -   [List all spaces in Studio](#list-all-spaces-in-studio)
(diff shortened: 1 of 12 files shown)


frkntrcnn (dismissed):
lgtm

DOCUMENTATION.md
```diff
@@ -9,6 +9,8 @@
         -   [Pull package for EMS Store](#pull-package-for-ems-store)
     -   [Pull/Push individual assets from/to Studio](#pullpush-individual-assets-fromto-studio)
     -   [Overwrite Package In Studio](#overwrite-package-in-studio)
+    -   [Export multiple packages at once from Studio](#export-multiple-packages-at-once-from-studio)
```
promeris, line 12: I think a more correct term would be batch export and import. Let's replace it everywhere in the docs.

src/util/semantic-versioning.ts
```diff
@@ -0,0 +1,23 @@
+export class SemanticVersioning {
+    private version: string;
+
+    constructor(version: string) {
+        this.version = version;
+    }
```
promeris, line 6: nit: Can be simplified to the following.
```suggestion
    constructor(private version: string) {
    }
```

fisgeci: Unfortunately this doesn't work. I tried it out as i wasn't sure but it throws a duplicate identifier compile error.

promeris (dismissed):
Looks good. If you are done with testing let's also bump the patch version.

promeris (approved):

PR #148, opened 2023-09-19
TN-4564: Improve performance of listing and export

fisgeci:
#### Description
The endpoint we were using for fetching datapools was really slow. We swapped it out for a more lightweight endpoint that returns less data. We also removed the dataPool name  and changed the signature of the listing and manifest file to comply to the new endpoint.
#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celonis/content-cli",
-  "version": "0.4.9",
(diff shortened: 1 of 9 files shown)


promeris (dismissed):

fisgeci:
Will merge this after the 25ft.

promeris (approved):

PR #152, opened 2023-10-05
tracing headers fix

frkntrcnn:
#### Description

Passing `x-datadog-parent-id` header as null causes a generation of new trace-id on the service layer. Populated `x-datadog-parent-id` via trace-id as well to avoid it.

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celonis/content-cli",
-  "version": "0.5.1",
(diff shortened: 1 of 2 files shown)


promeris (dismissed):

ashishkubaghel (dismissed):
LGTM

promeris (approved):

PR #154, opened 2023-11-24
[TA-2675] Add command for listing possible assignment values

LaberionAjvazi:
#### Description

Added command for listing possible variable assignment values.

[Design doc](https://docs.google.com/document/d/1gqn3CjE8rFPf_eLgimL55qwl_nGTjhfweiDmmaSFitE)

#### Checklist

### DIFF
--- a/src/api/variables-api.ts
+++ b/src/api/variables-api.ts
@@ -4,14 +4,49 @@ import {
 } from "../interfaces/package-manager.interfaces";
 import {httpClientV2} from "../services/http-client-service.v2";
 import {FatalError} from "../util/logger";
+import {variableAssignmentApis} from "./variable-assignment-apis";
+import {URLSearchParams} from "url";
 
 class VariablesApi {
     public static readonly INSTANCE = new VariablesApi();
+    public static readonly ASSIGNMENT_APIS  = variableAssignmentApis;
+
 
     public async assignVariableValues(packageKey: string, variablesAssignments: VariablesAssignments[]): Promise<ContentNodeTransport[]> {
         return httpClientV2.post(`/package-manager/api/nodes/by-package-key/${packageKey}/variables/values`, variablesAssignments).catch(e => {
             throw new FatalError(`Problem updating variables of package ${packageKey}: ${e}`);
         });
+    }
+
+    public async getCandidateAssignments(type: string, params?: string): Promise<Object[]> {
+        const apiUrl: string = VariablesApi.ASSIGNMENT_APIS[type].url;
+
+        if (apiUrl == null) {
+            throw new FatalError(`Variable type ${type} not supported.`);
+        }
+
+        const queryParams = this.parseParams(params);
+
+        return httpClientV2.get(apiUrl + `?${queryParams?.toString()}`).catch(e => {
+            throw new FatalError(`Problem getting variables assignment values for type ${type}: ${e}`);
+        });
+    }
+
+    private parseParams(params?: string): URLSearchParams {
+        const queryParams = new URLSearchParams();
+
+        if (params) {
+            try {
+                params.split(",").forEach((param: string) => {
+                    const paramKeyValuePair: string[] = param.split("=");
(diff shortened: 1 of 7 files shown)


src/api/variables-api.ts
```diff
@@ -4,15 +4,31 @@ import {
...
     }
+
+    public async getCandidateAssignments(type: string, params?: URLSearchParams): Promise<object[]> {
+        const apiUrl: string = VariablesApi.ASSIGNMENT_APIS[type].url;
+
+        if (apiUrl == null) {
+            throw new FatalError(`Variable type ${type} not supported.`);
+        }
+
+        return httpClientV2.get(apiUrl + `?${params?.toString()}`).catch(e => {
```
promeris, line 28: In case no params are passed, then it would send `?undefined`? Does that work properly?

src/commands/variable.command.ts
```diff
@@ -0,0 +1,12 @@
+import {variableService} from "../services/package-manager/variable-service";
+
+export class VariableCommand {
+
+    public async listAssignments(variableType: string, jsonResponse: boolean, params: string): Promise<void> {
+        if (jsonResponse) {
+            await variableService.listCandidateAssignments(variableType, params);
+        } else {
+            await variableService.findAndExportCandidateAssignments(variableType, params);
+        }
```
promeris, line 10: The nested statements should be the other way around, as it should export when `--json` flag is passed.

src/content-cli-list.ts
```diff
@@ -66,6 +67,22 @@ export class List {
...
         return program;
     }
+
+    public static assignments(program: CommanderStatic): CommanderStatic {
+        program
+            .command("assignments")
+            .description("Command to list possible variable assignments for a type")
+            .option("-p, --profile <profile>", "Profile which you want to use to list possible variable assignments")
+            .option("--json", "Return response as json type", "")
+            .option("--type <type>", "Type of variable")
```
promeris, line 77: This should be a required option, right?

src/api/variables-api.ts
ZgjimHaziri: nit: Not very familiar with this `httpClientV2`, but can you check if it has some other way of providing the params? Since there is a `URLSearchParams` object, I'd think is some URL object which has fields like: `url`, `params` etc.

LaberionAjvazi: `httpClientV2` is a custom service that accepts only a string param for the get URL so it does not have any additional fields that we can use.

ZgjimHaziri (dismissed):

promeris (dismissed):

gramozkaragjyzi (dismissed):

src/api/variables-api.ts
```diff
@@ -4,15 +4,35 @@ import {
...
+
 
     public async assignVariableValues(packageKey: string, variablesAssignments: VariablesAssignments[]): Promise<ContentNodeTransport[]> {
         return httpClientV2.post(`/package-manager/api/nodes/by-package-key/${packageKey}/variables/values`, variablesAssignments).catch(e => {
             throw new FatalError(`Problem updating variables of package ${packageKey}: ${e}`);
         });
     }
+
+    public async getCandidateAssignments(type: string, params: URLSearchParams): Promise<object[]> {
+        let apiUrl: string = VariablesApi.ASSIGNMENT_APIS[type]?.url;
```
AronHoxha, line 22: nit
```suggestion
        const apiUrl: string = VariablesApi.ASSIGNMENT_APIS[type]?.url;
```

AronHoxha: nvm I see it is being reassigned, though in that note I would suggest to try and keep it const ,so as to not change values of variables (arguably making it easier to separate code into pure functions in the future)

LaberionAjvazi: I inverted the conditions and switched the variable to const.

AronHoxha (approved):

ZgjimHaziri (approved):

PR #155, opened 2023-11-30
Added ability to update connection properties programmatically

celmk:
#### Description

Customer requirement to be able to programmatically update connection properties. Use case is connection properties or credentials being rotated automatically, and these need to be integrated into Celonis for compliance and governance reasons. This input can not be done manually by a user.

Command has been added in a way that supports adding updates to properties of other configuration objects if necessary.

#### Checklist

- [ X ] I have self-reviewed this PR
- [ X ] I have tested the change and proved that it works in different scenarios
- [ X ] I have updated docs if needed

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -17,6 +17,7 @@
 -   [Data Pool export / import commands](#data-pool-export--import-commands)
     - [Export Data Pool](#export-data-pool)
     - [Batch Import multiple Data Pools](#batch-import-multiple-data-pools)
+-   [Updating connection properties](#updating-connection-properties-programmatically)
 
 ## Content CLI Core Features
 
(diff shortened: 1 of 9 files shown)


src/services/connection/connection.service.ts
```diff
@@ -0,0 +1,56 @@
+import { dataPoolApi } from "../../api/data-pool-api";
+import { logger } from "../../util/logger";
+
+class ConnectionService {
+    public async findAllConnections(dataPoolId: string): Promise<any[]> {
+        const connections =  await dataPoolApi.listConnections(dataPoolId);
+        connections.forEach(connection => {
+            logger.info(`Connection ID: ${connection.id} - Name: ${connection.name} - Type: ${connection.type}`);
```
qmucolli, line 8: What's a typical number of connections returned here?
Wondering if this could clutter the logs and make it unreadable.

DOCUMENTATION.md
```diff
@@ -465,5 +466,24 @@ The command outputs an import report.
...
+```content-cli list connection --profile <profile> --dataPoolId <dataPoolId>```
+
+Depending on the type of source, the updatable properties may differ. You can use the following command to
+get a full list of properties for a data source connection:
+
+```content-cli get connection --profile <profile> --dataPoolId <dataPoolId> --connectionId <connectionId>```
+
+You can then update the property you want to update using the `set` command:
+
+```content-cli set connection --rpofile <profile> --dataPoolId <dataPoolId> --connectionId <connectionId> --property <property> --value <value>```
```
qmucolli, line 487:
```suggestion
```content-cli set connection --profile <profile> --dataPoolId <dataPoolId> --connectionId <connectionId> --property <property> --value <value>```
```

qmucolli (dismissed):

AlbinHoxha (dismissed):
Changes look good, but for these changes to show up for the clients you would have to bump the version in package.json

qmucolli (approved):

PR #160, opened 2023-12-07
[TA-2697] Add documentation for listing assignments/mapping of variables

LaberionAjvazi:
#### Description

Added documentation for changes merged in #153 and #154.
#### Checklist

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -13,6 +13,7 @@
     -   [Batch import packages into Studio](#importing-multiple-packages-into-studio)
     -   [List all spaces in Studio](#list-all-spaces-in-studio)
     -   [List all packages in Studio](#list-all-packages-in-studio)
+    -   [List assignments](#list-assignments)
     -   [Asset options for Analysis](#asset-options-for-analysis)
 -   [Data Pool export / import commands](#data-pool-export--import-commands)
     - [Export Data Pool](#export-data-pool)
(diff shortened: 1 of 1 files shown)


DOCUMENTATION.md
```diff
@@ -357,6 +358,28 @@ assets you have access to. 
...
+
+Currently, only variables of type DATA_MODEL and CONNECTION are supported.  
+
+The command includes an optional --params option for additional value filtering. 
+Parameters should be provided in CSV format. 
+For instance, when dealing with variables of type CONNECTION, you can use the --params option to retrieve 
+only the potential assignment values for connections with the appName 'Celonis' using the following command:
+
+```
+content-cli list assignments -p <your-chosen-profile> --variableType CONNECTION --params appName=Celonis
```
jetakasabaqi, line 380: This part looks good to me.

The previous PR which updated the [Batch Import packages into Studio](https://github.com/celonis/content-cli/blob/master/DOCUMENTATION.md#batch-import-packages-into-studio) I think it's no that clear.

We are saying that by default the variables are assigned, but that is not true. I think the wording should be that if the users have assigned the variables in the manifest file, then those values will be saved.

Also for assigning only the data models, I think we should mention that that is deprecated and will be removed in the future releases. What do you think?

LaberionAjvazi: Regarding the wording of the variable assignments, I think the second part of the sentence you mentioned clarifies the point you are making: `... as defined in the manifest.yml file of the exported packages.`.

As for the second point I agree, and added it to the doc.

jetakasabaqi (approved):

PR #161, opened 2023-12-08
[TA-2683] Setup tests and add tests for listing assignments

LaberionAjvazi:
#### Description

Setup testing with jest and add tests for list assignments command.

#### Checklist


jest.config.ts
```diff
@@ -0,0 +1,13 @@
+import type { Config } from "@jest/types"
+const config: Config.InitialOptions = {
+    verbose: true,
+    transform: {
+        "^.+\\.tsx?$": "ts-jest",
```
promeris, line 5: Do we need this configuration as there aren't any tsx files?

tests/utls/context-mock.ts
```diff
@@ -0,0 +1,12 @@
+import {contextService} from "../../src/services/context.service";
+
+export function setDefaultProfile(): void {
+    contextService.setContext({
+        profile: {
+            name: "test",
+            team: "https://myTeam.celonis.cloud/",
+            apiToken: "YnQ3N2M0M2ItYzQ3OS00YzgyLTg0ODgtOWNkNzhiNzYwOTU2OlFkNnBpVCs0M0JBYm1ZWGlCZ2hPd245aldwWTNubFQyYVFOTFBUeHEwdUxM",
```
promeris, line 8: I don't think we need a real token here :smile:.

tests/utls/context-mock.ts
```diff
@@ -0,0 +1,12 @@
+import {contextService} from "../../src/services/context.service";
+
+export function setDefaultProfile(): void {
```
qmucolli, line 3: Is this needed on a global level (should all tests have this by default)?

tests/utls/test-transport.ts
```diff
@@ -0,0 +1,15 @@
+import * as Transport from "winston-transport";
+import {LogEntry} from "winston";
+
+export class TestTransport extends Transport {
```
qmucolli, line 4: Could you elaborate what this is needed for? I'm not sure I'm following :D

tests/list/assignments.spec.ts
```diff
@@ -0,0 +1,81 @@
+import {VariableCommand} from "../../src/commands/variable.command";
+import axios from "axios";
+import {logger} from "../../src/util/logger";
+import {TestTransport} from "../utls/test-transport";
+import {setDefaultProfile} from "../utls/context-mock";
+import {FileService} from "../../src/services/file-service";
+import * as fs from "fs";
+import * as path from "path";
+
+jest.mock("axios");
```
qmucolli, line 10: I think these are better suited as part of a global setup instead of here, wdyt?

qmucolli (commented):
Generally looks good, I'm wondering if we should think of adding SonarCloud and also handle a global test config which would do the initial generation of each test?

jest.config.ts
LaberionAjvazi: `x` is optional in this expression. We need this to transform our `ts` files before running the tests.

tests/utls/context-mock.ts
LaberionAjvazi: This is a non-valid mocked token anyway.

tests/utls/context-mock.ts
LaberionAjvazi: Apart from tests for profile commands, I think all tests will need this.

tests/utls/test-transport.ts
LaberionAjvazi: I am using this to capture logs during testing. By attaching a custom transport (similar to [this](https://github.com/celonis/content-cli/blob/master/src/util/logger.ts#L24)) I can then read the logs as you would e.g. in Java by adding a TestAppender to logback.

tests/list/assignments.spec.ts
LaberionAjvazi: I agree and added a test setup file that handles mocking modules and the default profile.

promeris (approved):

promeris (dismissed):

qmucolli (dismissed):

promeris (dismissed):

promeris (approved):

PR #165, opened 2024-01-10
[TA-2747] Command for listing active packages across flavors

LaberionAjvazi:
#### Description

Added commands for listing active packages across flavors.

#### Checklist

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -13,13 +13,14 @@
     -   [Batch import packages into Studio](#importing-multiple-packages-into-studio)
     -   [List all spaces in Studio](#list-all-spaces-in-studio)
     -   [List all packages in Studio](#list-all-packages-in-studio)
-    -   [List all packages across flavors](#list-all-packages-across-flavors)
     -   [List assignments](#list-assignments)
     -   [Asset options for Analysis](#asset-options-for-analysis)
 -   [Data Pool export / import commands](#data-pool-export--import-commands)
     - [Export Data Pool](#export-data-pool)
     - [Batch Import multiple Data Pools](#batch-import-multiple-data-pools)
 -   [Updating connection properties](#updating-connection-properties-programmatically)
+-   [Config commands](#config-commands)
+    -   [List all packages across flavors](#list-all-packages-across-flavors)
 
 ## Content CLI Core Features
 
@@ -339,21 +340,6 @@ content-cli list packages -p <your-chosen-profile>
 content-cli list packages -p <profileName> --json --packageKeys <package1> <package2>
 ```
 
-### List all packages across flavors
-
-With this command you can retrieve a list of all packages across flavors within a team.
-The command takes your permissions into consideration and only lists the
-packages you have access to.
-
--   It is also possible to download packages in JSON format by adding '--json' option.
--   When the JSON format option is used, it is also possible to include package dependencies by adding '--withDependencies' flag.
--   When the JSON format option is used, it is also possible to filter packages by adding '--packageKeys' parameter
-
-```
-content-cli list exportPackages -p <your-chosen-profile>
-content-cli list exportPackages -p <profileName> --json --packageKeys <package1> <package2>
-```
-
 ### List all data pools of the team
 
 With this command you can retrieve a list of all data pools within a team.
@@ -527,5 +513,23 @@ get a full list of properties for a data source connection:
 You can then update the property you want to update using the `set` command:
 
 ```content-cli set connection --profile <profile> --dataPoolId <dataPoolId> --connectionId <connectionId> --property <property> --value <value>```
+
+### Config commands
+
+
+### List all packages across flavors
+
+With this command you can retrieve a list of all packages across flavors within a team.
+The command takes your permissions into consideration and only lists the
+packages you have access to.
+
+-   It is also possible to download packages in JSON format by adding '--json' option.
+-   When the JSON format option is used, it is also possible to include package dependencies by adding '--withDependencies' flag.
+-   When the JSON format option is used, it is also possible to filter packages by adding '--packageKeys' parameter
+
+```
+content-cli config list -p <your-chosen-profile>
+content-cli config list -p <profileName> --json --packageKeys <package1> <package2>
+```
 
 |--------------------------------------------------------------------------------------------------------------------------------|
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -522,6 +522,7 @@ You can then update the property you want to update using the `set` command:
 With this command you can retrieve a list of all packages across flavors within a team.
 The command takes your permissions into consideration and only lists the
 packages you have access to.
+To list only packages of certain flavors you can use the `--flavors` option and pass a list of flavors.
 
 -   It is also possible to download packages in JSON format by adding '--json' option.
 -   When the JSON format option is used, it is also possible to include package dependencies by adding '--withDependencies' flag.
@@ -529,7 +530,7 @@ packages you have access to.
 
 ```
 content-cli config list -p <your-chosen-profile>
-content-cli config list -p <profileName> --json --packageKeys <package1> <package2>
+content-cli config list -p <profileName> --json --packageKeys <package1> <package2> --flavors <flavor1> <flavor2>
 ```
 
 |--------------------------------------------------------------------------------------------------------------------------------|
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -19,8 +19,6 @@
     - [Export Data Pool](#export-data-pool)
     - [Batch Import multiple Data Pools](#batch-import-multiple-data-pools)
 -   [Updating connection properties](#updating-connection-properties-programmatically)
--   [Config commands](#config-commands)
-    -   [List all packages across flavors](#list-all-packages-across-flavors)
 
 ## Content CLI Core Features
 
@@ -513,24 +511,5 @@ get a full list of properties for a data source connection:
 You can then update the property you want to update using the `set` command:
 
 ```content-cli set connection --profile <profile> --dataPoolId <dataPoolId> --connectionId <connectionId> --property <property> --value <value>```
-
-### Config commands
-
-
-### List all packages across flavors
-
-With this command you can retrieve a list of all packages across flavors within a team.
-The command takes your permissions into consideration and only lists the
-packages you have access to.
-To list only packages of certain flavors you can use the `--flavors` option and pass a list of flavors.
-
--   It is also possible to download packages in JSON format by adding '--json' option.
--   When the JSON format option is used, it is also possible to include package dependencies by adding '--withDependencies' flag.
--   When the JSON format option is used, it is also possible to filter packages by adding '--packageKeys' parameter
-
-```
-content-cli config list -p <your-chosen-profile>
-content-cli config list -p <profileName> --json --packageKeys <package1> <package2> --flavors <flavor1> <flavor2>
-```
 
 |--------------------------------------------------------------------------------------------------------------------------------|
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -346,7 +346,7 @@ The command takes your permissions into consideration and only lists the
 packages you have access to.
 
 -   It is also possible to download packages in JSON format by adding '--json' option.
--   When the JSON format option is used, it is also possible to include package dependencies by adding '--includeDependencies' flag.
+-   When the JSON format option is used, it is also possible to include package dependencies by adding '--withDependencies' flag.
 -   When the JSON format option is used, it is also possible to filter packages by adding '--packageKeys' parameter
 
 ```
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -13,6 +13,7 @@
     -   [Batch import packages into Studio](#importing-multiple-packages-into-studio)
     -   [List all spaces in Studio](#list-all-spaces-in-studio)
     -   [List all packages in Studio](#list-all-packages-in-studio)
+    -   [List all packages across flavors](#list-all-packages-across-flavors)
     -   [List assignments](#list-assignments)
     -   [Asset options for Analysis](#asset-options-for-analysis)
 -   [Data Pool export / import commands](#data-pool-export--import-commands)
@@ -336,6 +337,21 @@ packages you have access to. 
 ```
 content-cli list packages -p <your-chosen-profile>
 content-cli list packages -p <profileName> --json --packageKeys <package1> <package2>
+```
+
+### List all packages across flavors
+
+With this command you can retrieve a list of all packages across flavors within a team.
+The command takes your permissions into consideration and only lists the
+packages you have access to.
+
+-   It is also possible to download packages in JSON format by adding '--json' option.
+-   When the JSON format option is used, it is also possible to include package dependencies by adding '--includeDependencies' flag.
+-   When the JSON format option is used, it is also possible to filter packages by adding '--packageKeys' parameter
+
+```
+content-cli list exportPackages -p <your-chosen-profile>
+content-cli list exportPackages -p <profileName> --json --packageKeys <package1> <package2>
 ```
 
 ### List all data pools of the team
--- a/src/api/batch-import-export-api.ts
+++ b/src/api/batch-import-export-api.ts
@@ -1,0 +1,31 @@
+import {PackageExportTransport} from "../interfaces/package-export-transport";
+import {httpClientV2} from "../services/http-client-service.v2";
+import {FatalError} from "../util/logger";
+
+class BatchImportExportApi {
+    public static readonly INSTANCE = new BatchImportExportApi();
+
+    public async findAllActivePackages(flavors: string[], withDependencies: boolean = false): Promise<PackageExportTransport[]> {
+        const queryParams = new URLSearchParams();
+
+        queryParams.set("withDependencies", withDependencies.toString());
+        flavors.forEach(flavor => queryParams.append("flavors", flavor))
+
+        return httpClientV2.get(`/package-manager/api/core/packages/export/list?${queryParams.toString()}`).catch(e => {
+            throw new FatalError(`Problem getting active packages: ${e}`);
+        });
+    }
+
+    public async findActivePackagesByKeys(packageKeys: string[], withDependencies: boolean = false): Promise<PackageExportTransport[]> {
+        const queryParams = new URLSearchParams();
+
+        packageKeys.forEach(key => queryParams.append("packageKeys", key))
+        queryParams.set("withDependencies", withDependencies.toString());
+
+        return httpClientV2.get(`/package-manager/api/core/packages/export/list-by-keys?${queryParams.toString()}`).catch(e => {
+            throw new FatalError(`Problem getting active packages by keys: ${e}`);
+        });
+    }
+}
+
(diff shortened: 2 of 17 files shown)


src/interfaces/package-export-transport.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export interface PackageExportTransport {
+    id: string;
+    key: string;
+    name: string;
+    changeDate: string;
+    activatedDraftId: string;
+    workingDraftId: string;
+    flavor: string;
+    version: PackageVersionTransport;
```
jetakasabaqi, line 23: With the latest [updates](https://github.com/celonis/cloud-package-manager/blob/develop/cloud-package-manager/src/main/java/cloud/celonis/packages/manager/exportimport/PackageExportTransport.java#L25) "version" is of type String.

src/services/package-manager/batch-import-export-service.ts
```diff
@@ -0,0 +1,59 @@
...
+            packagesToExport = await batchImportExportApi.findActivePackagesByKeys(packageKeys, withDependencies);
+        } else  {
+            packagesToExport = await batchImportExportApi.findAllActivePackages(flavors, withDependencies);
+        }
+
+        const packageByKey = new Map<string, PackageExportTransport>();
+        packagesToExport.forEach(pkg => packageByKey.set(pkg.key, pkg));
+
+        if (withDependencies) {
+            const studioPackageWithDataModelVariableAssignments = await variableService.getVariableAssignmentsForNodes(PackageManagerVariableType.DATA_MODEL);
```
jetakasabaqi, line 33: Do we know why the variable assignment for the data model, is dependent on the **withDependencies** flag?

src/services/package-manager/batch-import-export-service.ts
```diff
@@ -0,0 +1,59 @@
...
+            const dataModelDetailsByNode = await dataModelService.getDataModelDetailsForPackages(studioPackageWithDataModelVariableAssignments);
+            studioPackageWithDataModelVariableAssignments.forEach(studioPackage => {
```
jetakasabaqi, line 35: Do you think it's better if we extract these custom studio logic into a new service, and handle everything there? I think it would help us to keep the flow clean and to easily see what is being done for studio purposes.

src/services/package-manager/batch-import-export-service.ts
LaberionAjvazi: CDP is using this to get the data pools as package dependencies (together with other packages as dependencies) when resolving and selecting dependencies for copying during team-to-team-copy setup.

jetakasabaqi: ah yes, thanks!

src/services/package-manager/batch-import-export-service.ts
```diff
@@ -0,0 +1,59 @@
...
+                if (packageByKey.has(studioPackage.key)) {
+                    packageByKey.get(studioPackage.key).datamodels = dataModelDetailsByNode.get(studioPackage.key);
+                }
+            });
+        }
+
+        const studioPackages = await packageApi.findAllPackages();
```
jetakasabaqi, line 42: can we check if we can call by keys here, so we don't fetch the whole studio packages?

LaberionAjvazi: There is no API for getting a list of packages by keys. We can refactor the code to reuse packages with variables that we get [here](https://github.com/celonis/content-cli/pull/165/files#diff-10b781e4b0f7b649247dbfc2845e42dd87ef2b6ea9f130b055c3675655151d2dR25) or add a call to [get a single package with variables](https://www.celonis.dev/catalog/default/api/cloud-package-manager-package-manager-api/definition#/package/findPackageWithVariableAssignments). What do you think?

jetakasabaqi: We can reuse those then, by fetching them once and passing on the list in the other two dependent packages.

DOCUMENTATION.md
```diff
@@ -338,6 +339,21 @@ content-cli list packages -p <your-chosen-profile>
...
+With this command you can retrieve a list of all packages across flavors within a team.
+The command takes your permissions into consideration and only lists the
+packages you have access to.
+
+-   It is also possible to download packages in JSON format by adding '--json' option.
+-   When the JSON format option is used, it is also possible to include package dependencies by adding '--withDependencies' flag.
+-   When the JSON format option is used, it is also possible to filter packages by adding '--packageKeys' parameter
+
+```
+content-cli list exportPackages -p <your-chosen-profile>
```
jetakasabaqi, line 353: We were discussing with @promeris how should we name these new commands and were thinking that maybe we could do something like this:
```
content-cli config list 
content-cli config export
content-cli config import
```
What do you think?

DOCUMENTATION.md
```diff
@@ -512,4 +514,22 @@ You can then update the property you want to update using the `set` command:
 
 ```content-cli set connection --profile <profile> --dataPoolId <dataPoolId> --connectionId <connectionId> --property <property> --value <value>```
 
+### Config commands
+
+
+### List all packages across flavors
+
+With this command you can retrieve a list of all packages across flavors within a team.
```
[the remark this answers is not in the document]
LaberionAjvazi: For listing, it is possible to use the optional `--flavors` param to filter by flavors, but the default is to return all packages across flavors.
There are no restrictions regarding import/export, but if we want to put such restrictions, I think they belong on package-manager as the CLI, at its core, merely proxies to it.

src/api/batch-import-export-api.ts
```diff
@@ -0,0 +1,31 @@
+import {PackageExportTransport} from "../interfaces/package-export-transport";
+import {httpClientV2} from "../services/http-client-service.v2";
+import {FatalError} from "../util/logger";
+
+class BatchImportExportApi {
+    public static readonly INSTANCE = new BatchImportExportApi();
+
+    public async findAllActivePackages(flavors: string[], withDependencies: boolean = false): Promise<PackageExportTransport[]> {
```
AronHoxha, line 8: super-nit: It doesn't seem like these methods need to be async (await is not used)

src/api/batch-import-export-api.ts
```diff
@@ -0,0 +1,31 @@
...
+        const queryParams = new URLSearchParams();
+
+        queryParams.set("withDependencies", withDependencies.toString());
```
AronHoxha, line 11: Does the URLSearchParams constructor allow setting these values while constructing the object instead of assigning values? If yes, then it might be a good idea to utilize that instead to try and preserve referential transparency

src/interfaces/package-export-transport.ts
```diff
@@ -0,0 +1,20 @@
+import {StudioComputeNodeDescriptor} from "./package-manager.interfaces";
+
+export interface DependencyTransport {
```
AronHoxha, line 3: Are there any plans on using openapi  for generating these transports?

DOCUMENTATION.md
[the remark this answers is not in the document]
jetakasabaqi: All configurations can be exported from the API, as for the Team To Team copy feature it is up to their service to keep the information of which flavors are enabled and send those in the mentioned param.

src/commands/config.command.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export class ConfigCommand {
+
+    public async listActivePackages(jsonResponse: boolean, flavors: string[], withDependencies: boolean, packageKeys:string[]): Promise<void> {
+        if (jsonResponse) {
+            await batchImportExportService.findAndExportListOfActivePackages(flavors ?? [], packageKeys ?? [], withDependencies)
+        } else {
+            await batchImportExportService.listActivePackages(flavors ?? []);
+        }
+    }
```
AronHoxha, line 11: super-nit:
```suggestion
    public listActivePackages(jsonResponse: boolean, flavors: string[], withDependencies: boolean, packageKeys:string[]): Promise<void> {
    return jsonResponse 
      ? batchImportExportService.findAndExportListOfActivePackages(flavors ?? [], packageKeys ?? [], withDependencies)
      : batchImportExportService.listActivePackages(flavors ?? []);
    }
```

src/services/package-manager/batch-import-export-service.ts
```diff
@@ -0,0 +1,38 @@
...
+            packagesToExport = await batchImportExportApi.findAllActivePackages(flavors, withDependencies);
+        }
+
+        packagesToExport = await studioService.getExportPackagesWithStudioData(packagesToExport, withDependencies);
+
+        this.exportListOfPackages(packagesToExport);
+    }
+
+    private exportListOfPackages(packages: PackageExportTransport[]): void {
+        const filename = uuidv4() + ".json";
```
AronHoxha, line 32: could we use a more descriptive name for the json file instead of a UUID?

src/services/studio/studio.service.ts
```diff
@@ -0,0 +1,49 @@
...
+        await studioService.setSpaceIdForStudioPackages(packagesToExport, studioPackagesWithDataModels);
+
+        if (withDependencies) {
+            await studioService.setDataModelsForStudioPackages(packagesToExport, studioPackagesWithDataModels);
+        }
+
+        return packagesToExport;
+    }
+
+    private async setSpaceIdForStudioPackages(packages: PackageExportTransport[], studioPackages: PackageWithVariableAssignments[]): Promise<void> {
```
AronHoxha, line 20: does not need to be async or promise

src/services/studio/studio.service.ts
```diff
@@ -0,0 +1,49 @@
...
+        const packageByKey = new Map<string, PackageExportTransport>();
+        packages.forEach(pkg => packageByKey.set(pkg.key, pkg));
```
AronHoxha, line 22: could we make this a pure function by not changing the arguments supplied to the function and instead return new objects

src/services/studio/studio.service.ts
```diff
@@ -0,0 +1,49 @@
...
+
+        if (withDependencies) {
+            await studioService.setDataModelsForStudioPackages(packagesToExport, studioPackagesWithDataModels);
+        }
+
+        return packagesToExport;
+    }
+
+    private async setSpaceIdForStudioPackages(packages: PackageExportTransport[], studioPackages: PackageWithVariableAssignments[]): Promise<void> {
+        const packageByKey = new Map<string, PackageExportTransport>();
```
AronHoxha, line 21: from personal experience using an object (Record<string. PackageExportTransport) as a Map tends to be easier as well as more performant if the Map keys are simply keys

src/services/studio/studio.service.ts
```diff
@@ -0,0 +1,49 @@
+import {PackageExportTransport} from "../../interfaces/package-export-transport";
+import {packageApi} from "../../api/package-api";
+import {PackageManagerVariableType, PackageWithVariableAssignments} from "../../interfaces/package-manager.interfaces";
+import {dataModelService} from "../package-manager/datamodel-service";
+
+class StudioService {
+
+    public async getExportPackagesWithStudioData(packagesToExport: PackageExportTransport[], withDependencies: boolean) {
```
AronHoxha, line 8: nit: add return type

src/api/batch-import-export-api.ts
```diff
@@ -0,0 +1,31 @@
+import {PackageExportTransport} from "../interfaces/package-export-transport";
+import {httpClientV2} from "../services/http-client-service.v2";
+import {FatalError} from "../util/logger";
+
+class BatchImportExportApi {
+    public static readonly INSTANCE = new BatchImportExportApi();
```
AronHoxha, line 6: It feels to me like this is redundant, especially considering that the class is not exported

src/services/studio/studio.service.ts
```diff
@@ -0,0 +1,49 @@
...
+                packageByKey.get(studioPackage.key).spaceId = studioPackage.spaceId;
+            }
+        });
+    }
+
+    private async setDataModelsForStudioPackages(packages: PackageExportTransport[], studioPackageWithDataModels: PackageWithVariableAssignments[]): Promise<void> {
+        const packageByKey = new Map<string, PackageExportTransport>();
+        packages.forEach(pkg => packageByKey.set(pkg.key, pkg));
+
+        const dataModelDetailsByNode = await dataModelService.getDataModelDetailsForPackages(studioPackageWithDataModels);
```
AronHoxha, line 35: I think it would be clearer if dataModelDetailsByNode were sent as a function parameter instead, and this method be kept synchronous

src/api/batch-import-export-api.ts
LaberionAjvazi: This is the pattern that is used across the CLI. I think it is meant to provide singleton instances of classes.

src/interfaces/package-export-transport.ts
LaberionAjvazi: I am not aware of any plans to use OpenApi at the moment.

src/api/batch-import-export-api.ts
AronHoxha: Sure, but you are still just exporting the instance created from the class (the class is not usable outside this file), and considering the the constructor is not private there is nothing preventing you from creating multiple instances of the class.
This pattern did not seem to be used in the service classes for instance

src/services/package-manager/batch-import-export-service.ts
```diff
@@ -0,0 +1,38 @@
...
+        } else  {
+            packagesToExport = await batchImportExportApi.findAllActivePackages(flavors, withDependencies);
+        }
+
+        packagesToExport = await studioService.getExportPackagesWithStudioData(packagesToExport, withDependencies);
+
+        this.exportListOfPackages(packagesToExport);
+    }
+
+    private exportListOfPackages(packages: PackageExportTransport[]): void {
```
fatimarmullaku, line 31: Have you taken into consideration to actually export this as a ZIP file with each package having its own JSON file?
I'm not really clear on the feature yet, but having a long list of packages here can make it unreadable

src/commands/config.command.ts
LaberionAjvazi: I think its more readable with the if/else structure and also used in that way in other commands that have the json option as well.

DOCUMENTATION.md
```diff
@@ -512,4 +514,23 @@ You can then update the property you want to update using the `set` command:
...
+packages you have access to.
+To list only packages of certain flavors you can use the `--flavors` option and pass a list of flavors.
+
+-   It is also possible to download packages in JSON format by adding '--json' option.
+-   When the JSON format option is used, it is also possible to include package dependencies by adding '--withDependencies' flag.
+-   When the JSON format option is used, it is also possible to filter packages by adding '--packageKeys' parameter
+
+```
+content-cli config list -p <your-chosen-profile>
+content-cli config list -p <profileName> --json --packageKeys <package1> <package2> --flavors <flavor1> <flavor2>
```
fatimarmullaku, line 533: It seems that on the actual usage, flavors are comma-separated. Should we specify that on the documentation as well?

DOCUMENTATION.md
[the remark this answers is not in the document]
fatimarmullaku: One thing that still seems a bit weird is that we're exposing a technical detail such as flavor to end users. I don't know the actual usage of this API, but seeing how this can be used from end users, it might be worth considering to rename **flavors** to something else on the CLI side.

src/interfaces/package-export-transport.ts
AronHoxha: in that case I would suggest creating a ticket to have it documented as I would imagine that using openapi contract would be the ideal place to be for any client of the package-manager apis

src/services/package-manager/batch-import-export-service.ts
AronHoxha: In that case at the very list adding a specific static prefix would make the generated files a lot easier to read, for example packages-<uuid>

src/services/package-manager/batch-import-export-service.ts
LaberionAjvazi: No, we did not consider that. We wanted to make this similar to listing other resources (e.g. packages, data-pools) which are also exported as json.

src/services/package-manager/batch-import-export-service.ts
LaberionAjvazi: All other listing commands use randomized naming pattern. I would leave this as is and tackle them all at once if we decide to change this.

src/api/batch-import-export-api.ts
LaberionAjvazi: It does have some support, but instead of appending flavors we would have to join the list into a string (`flavors.join(",")` and allow it to be auto-parsed. I would prefer to leave it the way it is to prevent this auto-parsing, which might lead to unexpected behavior.

src/services/studio/studio.service.ts
LaberionAjvazi: Can you explain a bit what do you mean by 'if the Map keys are simply keys'? I thought using a map here would be better since we are doing key searches.

AronHoxha: an object is basically a map right where the key is a string and the value is anything for example:
```ts
const obj = {
  firstProperty: 5,
  'second-property': true
  thirdProperty: {
    a: 'b'
  }
}
```
And the common map functions can be represented as:
```ts
obj['firstProperty']; (map.get())
'second-property' in obj; (map.has())
obj['thirdProperty'] = {}
```

DOCUMENTATION.md
LaberionAjvazi: Using comma-separated values might work because it is treating the whole string as an entry, but we want users to abide by the documentation to match the intended use.

src/services/studio/studio.service.ts
LaberionAjvazi: I don't think the performance gains would be significant in our case and I find it more readable and consistent if we stick to maps since they are used throughout the CLI.

DOCUMENTATION.md
fatimarmullaku: Just to confirm, this works both ways right?

1. `--flavors studio process-workspace`
2. `--flavors studio,process-workspace`

DOCUMENTATION.md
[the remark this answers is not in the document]
jetakasabaqi: This API is intended for Team To Team Copy for now and we don't have a plan on whether it should have usage from other users, as it might happen that in the future it becomes internal. 

Because the main usage will be from TTC I think it's fine to have it in the transports as flavor, as for the documentation we can update to say all configurations but that also leaves it unclear for now, as the users will not know what configuration is or for which services you can get it. Did you have a suggestion in mind?

DOCUMENTATION.md
LaberionAjvazi: Yes, that is correct.

DOCUMENTATION.md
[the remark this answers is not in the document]
fatimarmullaku: I don't have any specific suggestions in mind, maybe something like **configurations** or **configurationTypes**.

> This API is intended for Team To Team Copy for now and we don't have a plan on whether it should have usage from other users, as it might happen that in the future it becomes internal.

To my knowledge, we don't have anything to mark commands on content CLI as internal only. This means that this will be visible to end users if i'm not mistaken.

AronHoxha (dismissed):

jetakasabaqi (approved):

ZgjimHaziri (approved):

PR #166, opened 2024-01-17
TN-5045: Add exclude action flows flag for batch export and import

fisgeci:
#### Description

Please include a description of the change and why the change was made.

#### Checklist

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -319,7 +319,7 @@ By default, you can not overwrite a package in the target team. To do this you c
 content-cli import packages -p <profileName> --file <exportedPackagesFile> --dataModelMappingsFile <dataModelMappingsFile> --overwrite
 ```
 
-If you want to not overwrite a package but not the Action Flows inside that package you can use the --excludeActionFlows flag
+If you want to overwrite a package but not the Action Flows inside that package you can use the --excludeActionFlows flag
 
 ```
 // Example usage of excludeActionFlows
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -278,6 +278,13 @@ You can use the --includeDependencies flag to also export the dependencies of th
 content-cli export packages -p <profileName> --packageKeys <package1> <package2> --includeDependencies
 ```
 
+If you don't want to export any Action Flows with the packages you can use --excludeActionFlows flag
+
+```
+//Batch export packages with excluded Action Flows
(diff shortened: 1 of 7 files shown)


DOCUMENTATION.md
```diff
@@ -312,6 +319,13 @@ By default, you can not overwrite a package in the target team. To do this you c
 content-cli import packages -p <profileName> --file <exportedPackagesFile> --dataModelMappingsFile <dataModelMappingsFile> --overwrite
 ```
 
+If you want to not overwrite a package but not the Action Flows inside that package you can use the --excludeActionFlows flag
```
LaberionAjvazi, line 322: Should this be `If you want` (affirmative) ?

src/api/package-api.ts
```diff
@@ -18,10 +18,13 @@ class PackageApi {
         });
     }
 
-    public async exportPackage(rootPackageKey: string, version?: string): Promise<Buffer> {
+    public async exportPackage(rootPackageKey: string, version?: string, excludeActionFlows?: boolean): Promise<Buffer> {
         const queryParams = new URLSearchParams();
         queryParams.set("newKey", rootPackageKey);
         queryParams.set("version", version ?? "");
+        if (excludeActionFlows) {
```
LaberionAjvazi, line 25: Can we skip this `if` and send the flag anyways?

src/content-cli-export.ts
```diff
@@ -13,8 +13,9 @@ export class Export {
             .option("-p, --profile <profile>", "Profile which you want to use to list packages")
             .requiredOption("--packageKeys <packageKeys...>", "Exports only given package keys")
             .option("--includeDependencies", "Include variables and dependencies", "")
+            .option("--excludeActionFlows", "Don't export action flows", "false")
```
LaberionAjvazi, line 16: Can we make sure to cast this to boolean or leave it as empty string to avoid evaluating to true for string `"false"`?

DOCUMENTATION.md
jetakasabaqi, line 322: nit: If you **want**

CDP service would have to send this new flag always, to fix the bug. Were they notified of the change?

fisgeci: I contacted with CDP team. After we merge this change. They will be able to upgrade the version of content-cli and adapt the changes on their end.

fisgeci:
Labeling as do not merge till backend changes reach prod.

LaberionAjvazi (approved):

ashishkubaghel (approved):

PR #168, opened 2024-01-26
[TA-2785] List variables command

LaberionAjvazi:
#### Description

Added command to list variables for packages by key and version.

The command can be used to list package variables for given keys and version. It has an optional --json param which can be used to export the variables as json instead of just listing in the console.

#### Checklist

### DIFF
--- a/src/commands/config.command.ts
+++ b/src/commands/config.command.ts
@@ -18,4 +18,8 @@ export class ConfigCommand {
             await variableService.listVariables(keysByVersion, keysByVersionFile);
         }
     }
+
+    public batchExportPackages(packageKeys: string[], withDependencies: boolean = false): Promise<void> {
+        return batchImportExportService.batchExportPackages(packageKeys, withDependencies);
+    }
 }
--- a/src/commands/config.command.ts
+++ b/src/commands/config.command.ts
@@ -1,4 +1,5 @@
 import {batchImportExportService} from "../services/package-manager/batch-import-export-service";
+import {variableService} from "../services/package-manager/variable-service";
 
 export class ConfigCommand {
 
@@ -7,6 +8,14 @@ export class ConfigCommand {
             await batchImportExportService.findAndExportListOfActivePackages(flavors ?? [], packageKeys ?? [], withDependencies)
         } else {
             await batchImportExportService.listActivePackages(flavors ?? []);
+        }
+    }
+
+    public async listVariables(jsonResponse: boolean, keysByVersion: string[], keysByVersionFile: string): Promise<void> {
+        if (jsonResponse) {
+            await variableService.exportVariables(keysByVersion, keysByVersionFile);
+        } else {
+            await variableService.listVariables(keysByVersion, keysByVersionFile);
         }
     }
 }
--- a/src/services/package-manager/variable-service.ts
+++ b/src/services/package-manager/variable-service.ts
@@ -71,24 +71,34 @@ class VariableService {
     }
 
     private async getVersionedVariablesByKeyVersionPairs(keysByVersion: string[], keysByVersionFile: string): Promise<VariableManifestTransport[]> {
+        const variablesExportRequest: PackageKeyAndVersionPair[] = await this.buildKeyVersionPairs(keysByVersion, keysByVersionFile);
+
+        const variableManifests = await batchImportExportApi.findVariablesWithValuesByPackageKeysAndVersion(variablesExportRequest);
+        return studioService.fixConnectionVariables(variableManifests);
+    }
+
+    private async buildKeyVersionPairs(keysByVersion: string[], keysByVersionFile: string): Promise<PackageKeyAndVersionPair[]> {
         let variablesExportRequest: PackageKeyAndVersionPair[] = [];
 
         if (keysByVersion.length) {
-            variablesExportRequest = keysByVersion.map(keyAndVersion => {
-                const keyAndVersionSplit = keyAndVersion.split(":");
-                return {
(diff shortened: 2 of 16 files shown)


src/services/package-manager/variable-service.ts
```diff
@@ -54,6 +69,33 @@ class VariableService {
 
         return queryParams;
     }
+
+    private async getVersionedVariablesByKeyVersionPairs(keysByVersion: string[], keysByVersionFile: string): Promise<VariableManifestTransport[]> {
+        let variablesExportRequest: PackageKeyAndVersionPair[] = [];
+
+        if (keysByVersion.length) {
+            variablesExportRequest = keysByVersion.map(keyAndVersion => {
```
jetakasabaqi, line 77: nit(😨): maybe we can seperate these if blocks into smaller methods.

jetakasabaqi (dismissed):

src/commands/config.command.ts
```diff
@@ -10,6 +11,14 @@ export class ConfigCommand {
         }
     }
 
+    public async listVariables(jsonResponse: boolean, keysByVersion: string[], keysByVersionFile: string): Promise<void> {
+        if (jsonResponse) {
+            await variableService.exportVariables(keysByVersion, keysByVersionFile);
```
promeris, line 16: I think this method name will be confusing when there is an actual export variables :smile:. Feel free to resolve and fix with variables export command.

promeris (approved):

src/commands/config.command.ts
jetakasabaqi: Hm, this is the PR for the variables export command. I think the confusion is that if jsonReponse is false you need to just list, otherwise you will export (as in export to a file)

jetakasabaqi: @LaberionAjvazi can you add a short example, of how the output is for both in the description?

ZgjimHaziri (approved):

src/commands/config.command.ts
LaberionAjvazi: We discussed with @promeris and @jetakasabaqi regarding the naming of commands and wanted to switch to using `list variables` instead of `listVariables`. However, this was not possible with the current setup as the commander could not parse this command format. We will try to tackle this issue together with other import/export commands before releasing.

jetakasabaqi (approved):

Buqeta (approved):

tests/config/config-list-variables.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+        expect(testTransport.logMessages[0].message).toContain(FileService.fileDownloadedMessage);
+
+        const expectedFileName = testTransport.logMessages[0].message.split(FileService.fileDownloadedMessage)[1];
+        expect(mockWriteFileSync).toHaveBeenCalledWith(path.resolve(process.cwd(), expectedFileName), JSON.stringify(fixedVariableManifests), {encoding: "utf-8"});
+
+        const variableExportRequest = parse(mockedPostRequestBodyByUrl.get("https://myTeam.celonis.cloud/package-manager/api/core/packages/export/batch/variables-with-assignments"));
+        expect(variableExportRequest).toEqual(packageKeyAndVersionPairs);
+    })
+
+    it("Should throw error if no mapping or file path is provided", async () => {
```
ksalihu, line 142: I think the test name should be `Should throw error if no mapping and no file path is provided`

src/services/package-manager/variable-service.ts
```diff
@@ -54,6 +69,43 @@ class VariableService {
...
+        const variableManifests = await batchImportExportApi.findVariablesWithValuesByPackageKeysAndVersion(variablesExportRequest);
+        return studioService.fixConnectionVariables(variableManifests);
+    }
+
+    private async buildKeyVersionPairs(keysByVersion: string[], keysByVersionFile: string): Promise<PackageKeyAndVersionPair[]> {
+        let variablesExportRequest: PackageKeyAndVersionPair[] = [];
+
+        if (keysByVersion.length) {
+            variablesExportRequest = this.buildKeyAndVersionPairsFromArrayInput(keysByVersion);
+        } else if (!keysByVersion.length && keysByVersionFile.length) {
```
ksalihu, line 85: i think `keysByVersionFile.length` should be `keysByVersionFile !== ""`, right?

src/services/package-manager/variable-service.ts
```diff
@@ -33,9 +36,21 @@ class VariableService {
...
+
+        variableManifests.forEach(variableManifest => {
+            logger.info(JSON.stringify(variableManifest));
+        });
+    }
+
+    public async exportVariables(keysByVersion: string[], keysByVersionFile: string): Promise<void> {
+        const variableManifests = await this.getVersionedVariablesByKeyVersionPairs(keysByVersion, keysByVersionFile);
+
+        this.exportToJson(variableManifests);
```
ksalihu, line 53: Maybe this is a convention for how we do this, but I think it would be easier to remove the responsibility of printing from these services, and handle that in the command classes, e.g. `ConfigCommand`.
That way we would not need to have two method that are essentially the same, only difference is how they are consumed. And it would be very simple to know which method of the service to call.

What do you think does it make sense to leave the export/logging out of the services, or is the effort to change the existing services just not worth it?

LaberionAjvazi: Logging is used on almost all services so I don't think it makes sense to work on it for now. I added it as a topic for improvement on this [doc](https://docs.google.com/document/d/1vvqThDSm09_Ej_adwIpKSYSxadBgZJt-jC3nT6I919o).

PR #169, opened 2024-01-31
[TA-2737] Batch import command

LaberionAjvazi:
#### Description

Added command to batch import packages. 

Note: Documentation will be added before release.

#### Checklist

### DIFF
--- a/src/content-cli-config.ts
+++ b/src/content-cli-config.ts
@@ -34,7 +34,7 @@ export class Config {
             .action(async cmd => {
                 await new ConfigCommand().listVariables(cmd.json, cmd.keysByVersion, cmd.keysByVersionFile);
                 process.exit();
-            })
+            });
 
         return program;
     }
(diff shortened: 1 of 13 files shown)


src/content-cli-config.ts
```diff
@@ -53,6 +53,21 @@ export class Config {
 
         return program;
     }
+
+    public static import(program: CommanderStatic): CommanderStatic {
+        program
+            .command("import")
+            .description("Command to import package configs")
+            .option("-p, --profile <profile>", "Profile which you want to use to list packages")
```
jetakasabaqi, line 61: to import packages?

I know you said you would tackle these maybe in a follow up PR, just commenting to not forget this.

src/services/package-manager/package-service.ts
```diff
@@ -270,12 +271,11 @@ class PackageService {
         return importedPackages.includes(version);
     }
 
-    private async createBodyForImport(filename: string): Promise<object> {
-        return {
-            formData: {
-                package: await fs.createReadStream(filename, {encoding: null})
-            },
-        }
+    private createBodyForImport(filename: string): FormData {
```
jetakasabaqi, line 274: Why did we change this?

LaberionAjvazi: The `postFile` method of the `httpService` was generalized as it was specific to this method, so the change was needed to adapt to the new implementation.

jetakasabaqi (approved):

ZgjimHaziri (approved):

promeris (approved):

Buqeta (approved):

PR #173, opened 2024-02-12
TA-2855: Fix packageKey extraction from zip name

LaberionAjvazi:
#### Description

Fixed logic for extracting package key from exported zip name for keys with multiple underscores (e.g. UUIDs).
Moved logic of deleting and updating files to separate loops to avoid losing references between the two operations.

#### Checklist

### DIFF
--- a/src/services/package-manager/batch-import-export-service.ts
+++ b/src/services/package-manager/batch-import-export-service.ts
@@ -59,10 +59,14 @@ class BatchImportExportService {
         exportedPackagesZip.addFile(BatchExportImportConstants.STUDIO_FILE_NAME, Buffer.from(stringify(studioData), "utf8"));
 
         exportedPackagesZip.getEntries().forEach(entry => {
-            if (entry.name.endsWith(".zip") && studioPackageKeys.includes(entry.name.split("_")[0])) {
-                const updatedPackage = studioService.processPackageForExport(entry, exportedVariables);
+            if (entry.name.endsWith(".zip")) {
+                const lastUnderscoreIndex = entry.name.lastIndexOf("_");
+                const namePart = entry.name.substring(0, lastUnderscoreIndex);
 
-                exportedPackagesZip.updateFile(entry, updatedPackage.toBuffer());
+                if (studioPackageKeys.includes(namePart)) {
+                    const updatedPackage = studioService.processPackageForExport(entry, exportedVariables);
+                    exportedPackagesZip.updateFile(entry, updatedPackage.toBuffer());
+                }
             }
         });
 
--- a/src/services/package-manager/batch-import-export-service.ts
+++ b/src/services/package-manager/batch-import-export-service.ts
@@ -61,9 +61,9 @@ class BatchImportExportService {
         exportedPackagesZip.getEntries().forEach(entry => {
             if (entry.name.endsWith(".zip")) {
                 const lastUnderscoreIndex = entry.name.lastIndexOf("_");
-                const namePart = entry.name.substring(0, lastUnderscoreIndex);
+                const packageKey = entry.name.substring(0, lastUnderscoreIndex);
 
-                if (studioPackageKeys.includes(namePart)) {
+                if (studioPackageKeys.includes(packageKey)) {
                     const updatedPackage = studioService.processPackageForExport(entry, exportedVariables);
                     exportedPackagesZip.updateFile(entry, updatedPackage.toBuffer());
                 }
--- a/src/services/studio/studio.service.ts
+++ b/src/services/studio/studio.service.ts
@@ -125,38 +125,35 @@ class StudioService {
     }
 
     private deleteScenarioAssets(packageZip: AdmZip): void {
-        packageZip.getEntries().forEach(entry => {
-            if (entry.entryName.startsWith("nodes/") && entry.entryName.endsWith(".yml")) {
+        packageZip.getEntries().filter(entry => entry.entryName.startsWith("nodes/") && entry.entryName.endsWith(".yml"))
+            .forEach(entry => {
                 const node: NodeExportTransport = parse(entry.getData().toString());
                 if (node.type === "SCENARIO") {
                     packageZip.deleteFile(entry);
                 }
-            }
-        })
+            });
     }
 
     private fixConnectionVariablesForRootNodeFiles(packageZip: AdmZip, zipName: string, exportedVariables: VariableManifestTransport[]): void {
-        packageZip.getEntries().forEach(entry => {
-            if (entry.name === "package.yml") {
-                const packageKeyAndVersion = this.getPackageKeyAndVersion(zipName);
+        const packageKeyAndVersion = this.getPackageKeyAndVersion(zipName);
 
-                const connectionVariablesByKey = this.getConnectionVariablesByKeyForPackage(packageKeyAndVersion.packageKey, packageKeyAndVersion.version, exportedVariables);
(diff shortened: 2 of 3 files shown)


src/services/package-manager/batch-import-export-service.ts
```diff
@@ -59,10 +59,14 @@ class BatchImportExportService {
         exportedPackagesZip.addFile(BatchExportImportConstants.STUDIO_FILE_NAME, Buffer.from(stringify(studioData), "utf8"));
 
         exportedPackagesZip.getEntries().forEach(entry => {
-            if (entry.name.endsWith(".zip") && studioPackageKeys.includes(entry.name.split("_")[0])) {
-                const updatedPackage = studioService.processPackageForExport(entry, exportedVariables);
-
-                exportedPackagesZip.updateFile(entry, updatedPackage.toBuffer());
+            if (entry.name.endsWith(".zip")) {
+                const lastUnderscoreIndex = entry.name.lastIndexOf("_");
+                const namePart = entry.name.substring(0, lastUnderscoreIndex);
```
jetakasabaqi, line 64: packageKey is more descriptive.

src/services/studio/studio.service.ts
```diff
@@ -125,34 +123,41 @@ class StudioService {
...
-        if (entry.name === "package.yml") {
-            const packageKeyAndVersion = zipName.replace(".zip", "").split("_");
-            const connectionVariablesByKey = this.getConnectionVariablesByKeyForPackage(packageKeyAndVersion[0], packageKeyAndVersion[1], exportedVariables);
+    private fixConnectionVariablesForRootNodeFiles(packageZip: AdmZip, zipName: string, exportedVariables: VariableManifestTransport[]): void {
+        packageZip.getEntries().forEach(entry => {
+            if (entry.name === "package.yml") {
+                const lastUnderscoreIndex = zipName.lastIndexOf("_");
+                const packageKey = zipName.replace(".zip", "").substring(0, lastUnderscoreIndex);
+                const packageVersion = zipName.replace(".zip", "").substring(lastUnderscoreIndex + 1);
+
```
jetakasabaqi, line 143: Maybe we can extract this into a small function that would return
```
/// replace and substring logic
    return {
        key: packageKey,
        version: packageVersion
    };

```

tests/config/config-export.spec.ts
```diff
@@ -393,6 +393,87 @@ describe("Config export", () => {
         expect(secondPackageContent.variables).toEqual([...secondPackageVariableDefinition]);
     })
 
+    it("Should export with SCENARIO nodes removed and CONNECTION variables fixed for package key with multiple underscores", async () => {
+        const manifest: PackageManifestTransport[] = [];
```
jetakasabaqi, line 397: Do we have a test for types of keys that don't have any underscores?

LaberionAjvazi: Yes, all other tests are with keys without underscores.

promeris (dismissed):

jetakasabaqi (dismissed):

Buqeta (dismissed):

src/services/studio/studio.service.ts
```diff
@@ -125,33 +124,49 @@ class StudioService {
...
+        })
     }
 
-    private fixConnectionVariablesIfRootNodeFile(packageZip: AdmZip, entry: IZipEntry, zipName: string, exportedVariables: VariableManifestTransport[]): void {
-        if (entry.name === "package.yml") {
-            const packageKeyAndVersion = zipName.replace(".zip", "").split("_");
-            const connectionVariablesByKey = this.getConnectionVariablesByKeyForPackage(packageKeyAndVersion[0], packageKeyAndVersion[1], exportedVariables);
+    private fixConnectionVariablesForRootNodeFiles(packageZip: AdmZip, zipName: string, exportedVariables: VariableManifestTransport[]): void {
+        packageZip.getEntries().forEach(entry => {
+            if (entry.name === "package.yml") {
```
ksalihu, line 140: I think we could use `find` here to find the `package.yml` file, and then perform the rest of the changes.

src/services/studio/studio.service.ts
```diff
@@ -125,33 +124,49 @@ class StudioService {
...
-                const exportedNode: NodeExportTransport = parse(entry.getData().toString());
-                const nodeContent: NodeSerializedContent = parse(exportedNode.serializedContent);
+                const connectionVariablesByKey = this.getConnectionVariablesByKeyForPackage(packageKeyAndVersion.packageKey, packageKeyAndVersion.version, exportedVariables);
 
-                nodeContent.variables = nodeContent.variables.map(variable => ({
-                    ...variable,
-                    metadata: variable.type === PackageManagerVariableType.CONNECTION ?
-                        connectionVariablesByKey.get(variable.key).metadata : variable.metadata
-                }));
+                if (connectionVariablesByKey.size) {
```
ksalihu, line 145: I think this part can be moved outside of the loop/find, and that way we can avoid finding the file in the first place, if we are not going to make any changes

src/services/studio/studio.service.ts
```diff
@@ -125,33 +124,49 @@ class StudioService {
...
     }
 
-    private deleteFileIfTypeScenario(packageZip: AdmZip, entry: IZipEntry): void {
-        if (entry.entryName.startsWith("nodes/") && entry.entryName.endsWith(".yml")) {
-            const node: NodeExportTransport = parse(entry.getData().toString());
-            if (node.type === "SCENARIO") {
-                packageZip.deleteFile(entry);
+    private deleteScenarioAssets(packageZip: AdmZip): void {
+        packageZip.getEntries().forEach(entry => {
+            if (entry.entryName.startsWith("nodes/") && entry.entryName.endsWith(".yml")) {
```
ksalihu, line 129: I think we can replace this with the `filter` method

src/services/studio/studio.service.ts
LaberionAjvazi: Since this entire service is not very 'pure function oriented' I moved the part where we read the file instead the if instead of extracting to a function. I think this is more readable than passing the same arguments in one more function since we are using the same references due to how AdmZip works. Do you think this is ok?

src/services/studio/studio.service.ts
```diff
@@ -125,33 +124,46 @@ class StudioService {
...
-                const exportedNode: NodeExportTransport = parse(entry.getData().toString());
-                const nodeContent: NodeSerializedContent = parse(exportedNode.serializedContent);
+        const connectionVariablesByKey = this.getConnectionVariablesByKeyForPackage(packageKeyAndVersion.packageKey, packageKeyAndVersion.version, exportedVariables);
 
-                nodeContent.variables = nodeContent.variables.map(variable => ({
-                    ...variable,
-                    metadata: variable.type === PackageManagerVariableType.CONNECTION ?
-                        connectionVariablesByKey.get(variable.key).metadata : variable.metadata
-                }));
+        if (connectionVariablesByKey.size) {
```
ksalihu, line 142: I would prefer this to be a number comparison.
` if(connectionVariablesByKey.size !== 0) {`
Alternatively, this could also be an early return:
```
if(connectionVariablesByKey.size === 0) {
    return;
}
```

ksalihu (dismissed):

src/services/studio/studio.service.ts
ksalihu: I meant to make the size check before doing the rest of the logic of the function. Which you updated in the subsequent commit. Marking this as resolved.

ksalihu (approved):

jetakasabaqi (approved):

PR #175, opened 2024-02-15
TA-2859: Move to space fails because of non existing space reference

jetakasabaqi:
#### Description
Previously we were importing the nodes with the source team Space ID set inside the files, this resulted in failing when trying to move the packages to the desired space in the target team, because of the non-existing space ID reference.

Changes in this PR include:

- Map spaces before doing an import
    - For each studio package, we will find the desired space ID in the target team
        - If the studio manifest file contains the space ID - (users have mapped it in the UI), that will be assigned
        - Else if we can find a space in the target with the same name, that will be assigned
        - Otherwise, we will create a new space.

Added test for the specific cases:
- Space ID in manifest file -> find by space ID 
- Space ID in the manifest file but does not exist in target team -> fails
- Space ID not provided in the manifest file, but space name yes -> finds it by name
- Space ID not provided in manifest file & not found any matching spaces with name -> create new space 

Didn't provide tests that will check if the imported files have the correct ID set as that requires unzipping each file. This part will be tackled with manual E2E tests.

#### Checklist

### DIFF
--- a/src/services/package-manager/batch-import-export-service.ts
+++ b/src/services/package-manager/batch-import-export-service.ts
@@ -15,6 +15,7 @@ import * as fs from "fs";
 import * as FormData from "form-data";
 import {BatchExportImportConstants} from "../../interfaces/batch-export-import-constants";
 import AdmZip = require("adm-zip");
+import {packageApi} from "../../api/package-api";
 
 class BatchImportExportService {
 
@@ -81,14 +82,14 @@ class BatchImportExportService {
         const configs = new AdmZip(file);
         const studioManifests: StudioPackageManifest[] = this.parseEntryData(configs, BatchExportImportConstants.STUDIO_FILE_NAME) as StudioPackageManifest[];
         const variablesManifest = this.parseEntryData(configs, BatchExportImportConstants.VARIABLES_FILE_NAME) as string;
-
+        const existingStudioPackages = await packageApi.findAllPackages();
 
         const updatedFiles = await studioService.mapSpaces(configs, studioManifests);
         const rootDirectoryZipPath= this.rezipRoot(updatedFiles);
         const formData = this.buildBodyForImport(rootDirectoryZipPath, variablesManifest);
 
         const postPackageImportData = await batchImportExportApi.importPackages(formData, overwrite);
-        await studioService.processImportedPackages(updatedFiles, [], studioManifests);
+        await studioService.processImportedPackages(updatedFiles, existingStudioPackages, studioManifests);
 
         const reportFileName = "config_import_report_" + uuidv4() + ".json";
         fileService.writeToFileWithGivenName(JSON.stringify(postPackageImportData), reportFileName);
--- a/src/services/package-manager/batch-import-export-service.ts
+++ b/src/services/package-manager/batch-import-export-service.ts
@@ -61,7 +61,7 @@ class BatchImportExportService {
         exportedPackagesZip.addFile(BatchExportImportConstants.STUDIO_FILE_NAME, Buffer.from(stringify(studioData), "utf8"));
 
         exportedPackagesZip.getEntries().forEach(entry => {
-            if (entry.name.endsWith(".zip")) {
+            if (entry.name.endsWith(BatchExportImportConstants.ZIP_EXTENSION)) {
                 const lastUnderscoreIndex = entry.name.lastIndexOf("_");
                 const packageKey = entry.name.substring(0, lastUnderscoreIndex);
 
--- a/src/services/package-manager/batch-import-export-service.ts
+++ b/src/services/package-manager/batch-import-export-service.ts
@@ -4,18 +4,17 @@ import {v4 as uuidv4} from "uuid";
 import {
     PackageExportTransport,
     PackageKeyAndVersionPair,
-    PackageManifestTransport, StudioPackageManifest,
+    PackageManifestTransport,
+    StudioPackageManifest,
     VariableManifestTransport
 } from "../../interfaces/package-export-transport";
 import {FileService, fileService} from "../file-service";
 import {studioService} from "../studio/studio.service";
 import {parse, stringify} from "../../util/yaml"
-import AdmZip = require("adm-zip");
 import * as fs from "fs";
 import * as FormData from "form-data";
 import {BatchExportImportConstants} from "../../interfaces/batch-export-import-constants";
-import {packageApi} from "../../api/package-api";
-import {Readable} from "stream";
+import AdmZip = require("adm-zip");
 
 class BatchImportExportService {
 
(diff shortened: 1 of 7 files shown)


src/services/studio/studio.service.ts
```diff
@@ -181,24 +180,69 @@ class StudioService {
...
+                    if(this.isStudioPackage(studioManifests, packageKey)) {
+                        const studioManifest = studioManifests.find(manifest => manifest.packageKey === packageKey);
+                        if (studioManifest) {
+
+                            const spaceId = await this.findDesiredSpaceIdForPackage(packageKey, studioManifest);
+                            studioManifest.space.id = spaceId;
+
+                            const packageZip = new AdmZip(packageZipFile.getData());
+                            packageZip.getEntries().forEach(nodeFile => {
+                                if (nodeFile.entryName.endsWith("yml")) {
```
promeris, line 196: Do we have constants for these filename extensions? Same for `.zip`

src/services/studio/studio.service.ts
```diff
@@ -181,24 +180,69 @@ class StudioService {
...
+            if(studioPackageManifest.space.id) {
+                const targetSpace = allSpaces.find(space => space.id === studioPackageManifest.space.id);
+                if (!targetSpace) {
+                    throw Error("Provided space ID does not exist.");
+                }
+                return targetSpace.id;
+            }
+
+            const targetSpaceByName = allSpaces.find(space => space.name === studioPackageManifest.space.name);
+            if(targetSpaceByName) {
```
promeris, line 225: I see in we are not following the same linting style in all written code. Could we please fix those?

src/services/studio/studio.service.ts
```diff
@@ -181,24 +180,69 @@ class StudioService {
...
+                }
+        }
+        return exportedFiles;
+    }
+
+    private isStudioPackage(studioManifests: StudioPackageManifest[], packageKey: string): boolean {
+        return studioManifests.some(manifest => manifest.packageKey === packageKey);
+    }
 
+    private async findDesiredSpaceIdForPackage(packageKey: string, studioPackageManifest: StudioPackageManifest): Promise<string> {
```
Buqeta, line 213: `packageKey` is not being used anywhere, so it can be omitted.

src/services/studio/studio.service.ts
```diff
@@ -80,17 +83,14 @@ class StudioService {
...
 
-    public async processImportedPackages(configs: AdmZip): Promise<void> {
-        const studioFile = configs.getEntry(BatchExportImportConstants.STUDIO_FILE_NAME);
-
-        if (studioFile) {
-            const studioManifests: StudioPackageManifest[] = parse(configs.getEntry(BatchExportImportConstants.STUDIO_FILE_NAME).getData().toString());
-
-            for (const manifest of studioManifests) {
-                await this.movePackageToSpace(manifest);
+    public async processImportedPackages(configs: AdmZip, existingStudioPackages: ContentNodeTransport[], studioManifests: StudioPackageManifest[]): Promise<void> {
```
Buqeta, line 86: `configs` is also not being used

src/services/studio/studio.service.ts
```diff
@@ -80,17 +83,14 @@ class StudioService {
...
+            for (const  manifest of studioManifests) {
+                const existingPackage = existingStudioPackages.find(existingPackage => existingPackage.key === manifest.packageKey);
+                if(existingPackage) {
+                    await packageApi.movePackageToSpace(existingPackage.id, manifest.space.id);
+                }
```
Buqeta, line 91: Can you fix the indentation?

src/services/studio/studio.service.ts
jetakasabaqi: added them.

src/services/package-manager/batch-import-export-service.ts
```diff
@@ -78,15 +80,29 @@ class BatchImportExportService {
 
     public async batchImportPackages(file: string, overwrite: boolean): Promise<void> {
         const configs = new AdmZip(file);
+        const studioManifests: StudioPackageManifest[] = this.parseEntryData(configs, BatchExportImportConstants.STUDIO_FILE_NAME) as StudioPackageManifest[];
+        const variablesManifest = this.parseEntryData(configs, BatchExportImportConstants.VARIABLES_FILE_NAME) as string;
```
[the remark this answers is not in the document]
jetakasabaqi: I guess it got left that way because we didn't need to do anything with it, but I mapped it into its respective type.

src/services/studio/studio.service.ts
jetakasabaqi: Did some changes with the latest commits, can you re-review?

LaberionAjvazi (dismissed):

src/services/package-manager/batch-import-export-service.ts
```diff
@@ -118,20 +124,37 @@ class BatchImportExportService {
         return batchImportExportApi.findVariablesWithValuesByPackageKeysAndVersion(variableExportRequest)
     }
 
-    private buildBodyForImport(file: string, configs: AdmZip): FormData {
+    private buildBodyForImport(configs: AdmZip, variablesManifests: VariableManifestTransport[]): FormData {
         const formData = new FormData();
+        const readableStream = this.getReadableStream(configs);
 
-        formData.append("file", fs.createReadStream(file));
+        formData.append("file", readableStream, {filename: "configs.zip"});
```
promeris, line 131: I'm curious on why the `filename` is required?

src/services/package-manager/batch-import-export-service.ts
```diff
@@ -5,15 +5,17 @@ import {
...
 import {FileService, fileService} from "../file-service";
 import {studioService} from "../studio/studio.service";
 import {parse, stringify} from "../../util/yaml"
-import AdmZip = require("adm-zip");
-import * as fs from "fs";
 import * as FormData from "form-data";
 import {BatchExportImportConstants} from "../../interfaces/batch-export-import-constants";
+import {packageApi} from "../../api/package-api";
+import {Readable} from "stream";
+import AdmZip = require("adm-zip");
```
ksalihu, line 18: q: Why is this using `require` instead of `import`?

promeris (dismissed):

src/services/package-manager/batch-import-export-service.ts
jetakasabaqi: When I try without it, this [error](https://celonis.datadoghq.com/logs?query=892c531d-a9be-41d1-bdcb-c34f43008a15%20&cols=core_host%2Ccore_service&event=AgAAAY3Q752c3NTq7AAAAAAAAAAYAAAAAEFZM1E3NmJWQUFDTkx6NDRkeURYWXdBbQAAACQAAAAAMDE4ZGQwZWYtYzgzNS00YzA1LTk1ZDktNWYwYjAwNzUxZTY0&index=%2A&messageDisplay=inline&stream_sort=desc&viz=stream) is being thrown

src/services/studio/studio.service.ts
```diff
@@ -181,24 +182,72 @@ class StudioService {
...
     private async assignRuntimeVariables(manifest: StudioPackageManifest): Promise<void> {
         if (manifest.runtimeVariableAssignments.length) {
             await variableService.assignVariableValues(manifest.packageKey, manifest.runtimeVariableAssignments);
         }
     }
+    private updateSpaceIdForNode(nodeContent: string, spaceId: string): string {
+        const exportedNode: NodeExportTransport = parse(nodeContent);
+        const oldSpaceId = exportedNode.unversionedMetadata.spaceId;
+
+        nodeContent = nodeContent.replace(new RegExp(oldSpaceId, "g"), spaceId);
```
ksalihu, line 248: q: Is there a reason to use a regex instead of `replaceAll` here?

tests/config/config-import.spec.ts
```diff
@@ -206,12 +315,12 @@ describe("Config import", () => {
...
         await new ConfigCommand().batchImportPackages("./export_file.zip", true);
 
         const expectedFileName = testTransport.logMessages[0].message.split(LOG_MESSAGE)[1];
         expect(mockWriteFileSync).toHaveBeenCalledWith(path.resolve(process.cwd(), expectedFileName), JSON.stringify(importResponse), {encoding: "utf-8"});
 
-        expect(mockedPostRequestBodyByUrl.get(assignVariablesUrl)).toEqual(JSON.stringify([variableAssignment]));
-    })
-})
\ No newline at end of file
+        expect(mockedPostRequestBodyByUrl.get(assignVariablesUrl)).toEqual(JSON.stringify([variableAssignment]));    })
```
ksalihu, line 325: I think we should write assertions on the `axios.post` spy instead

src/services/studio/studio.service.ts
jetakasabaqi: replaceAll is available from version 15 and the min required version to use Content CLI is >10, if we were to update this to 15 it requires announcing it as a change to all the user (studio personas).

src/services/package-manager/batch-import-export-service.ts
jetakasabaqi: should be from auto-import, will fix it.

tests/config/config-import.spec.ts
LaberionAjvazi: Assertions on `mockedPostRequestBodyByUrl` are the same since they only exist when the mock implementation is called ([ref](https://github.com/celonis/content-cli/blob/c5ca8e034d75753c63946b9556fec5abe5e48078/tests/utls/http-requests-mock.ts#L38)).

ksalihu: Sure, we can consider in the future if we want to improve this, since there are existing tests with this approach, I'm fine with leaving it this way

ksalihu (approved):

promeris (approved):

ZgjimHaziri (approved):

PR #182, opened 2024-04-17
TA-2944: Use hpagent when proxying requests

LaberionAjvazi:
#### Description

Axios fails to proxy requests from a http proxy to https servers (https://github.com/axios/axios/issues/4531) . This PR adds hpagent when proxying requests. 

#### Checklist

### DIFF
--- a/src/util/axios-initializer.ts
+++ b/src/util/axios-initializer.ts
@@ -1,0 +1,21 @@
+import axios, { AxiosInstance } from "axios";
+import { HttpsProxyAgent } from "hpagent";
+
+export class AxiosInitializer {
+    public static initializeAxios(): AxiosInstance {
+        const httpsProxy = process.env.https_proxy || process.env.HTTPS_PROXY;
+
+        if (httpsProxy) {
+            const httpsAgent = new HttpsProxyAgent({
+                proxy: httpsProxy
+            })
+
+            return axios.create({
+                httpsAgent,
+                proxy: false
+            })
+        } else {
+            return axios.create();
+        }
+    }
+}
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celonis/content-cli",
-  "version": "0.8.1",
+  "version": "0.8.2",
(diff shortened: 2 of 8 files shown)


src/util/axios-initializer.ts
```diff
@@ -0,0 +1,21 @@
+import axios, { AxiosInstance } from "axios";
+import { HttpsProxyAgent } from "hpagent";
+
+export class AxiosInitializer {
+    public static initializeAxios(): AxiosInstance {
+        const httpsProxy = process.env.https_proxy || process.env.HTTPS_PROXY;
+
+        if (httpsProxy) {
+            const httpsAgent = new HttpsProxyAgent({
+                proxy: httpsProxy
```
ZgjimHaziri, line 10: Did we look into the default values of the other properties here? I'd like to know specifically for this `keepAlive` value that's shown in the documentation.

LaberionAjvazi: No, I didn't look specifically into it, but the keepAlive is enabled by default. We can disable it or use another agent library like [https-proxy-agent](https://www.npmjs.com/package/https-proxy-agent), which does not enable it by default and seems to have a significantly more usage (downloads).

ZgjimHaziri: I see. I just wouldn't like to do any change here from how it was previously working, since that would need a more detailed assessment. If Axios/the previous request module was keeping the connection alive, we'd leave it like this.

LaberionAjvazi: Based on the [documentation](https://github.com/request/request/blob/3c0cddc7c8eb60b470e9519da85896ed7ee0081e/README.md) from the previous request module, it also defaulted to keep-alive:true by using a library called forever-agent, although I did not see any keepAliveInMs property to specify for how long.

ZgjimHaziri: Alright, the defaults look good to me then.

ZgjimHaziri (approved):

jetakasabaqi (approved):

Buqeta (approved):

ksalihu (approved):

PR #183, opened 2024-05-20
TA-2977: Rename config export properties to match API

promeris:
#### Description

Changes are made to ensure that used properties match the provided API interface.

#### Checklist

### DIFF
--- a/src/interfaces/package-export-transport.ts
+++ b/src/interfaces/package-export-transport.ts
@@ -51,7 +51,7 @@ export interface NodeExportTransport {
     name: string;
     type: string;
     exportSerializationType: string;
-    serializedContent: string;
+    configuration: string;
     schemaVersion: number;
 
     spaceId: string;
--- a/src/interfaces/package-export-transport.ts
+++ b/src/interfaces/package-export-transport.ts
@@ -48,18 +48,16 @@ export interface PackageKeyAndVersionPair {
 export interface NodeExportTransport {
     key: string;
     parentNodeKey: string;
-    packageNodeKey: string;
     name: string;
     type: string;
     exportSerializationType: string;
     serializedContent: string;
     schemaVersion: number;
 
-    unversionedMetadata: any;
-    versionedMetdata: object;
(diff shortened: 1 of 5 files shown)


src/interfaces/package-export-transport.ts
```diff
@@ -48,18 +48,16 @@ export interface PackageKeyAndVersionPair {
 export interface NodeExportTransport {
     key: string;
     parentNodeKey: string;
-    packageNodeKey: string;
```
ksalihu, line 51: Why did we remove the package node key?

ksalihu (dismissed):

src/interfaces/package-export-transport.ts
```diff
@@ -48,18 +48,16 @@ export interface PackageKeyAndVersionPair {
 export interface NodeExportTransport {
```
ksalihu, line 48: Should we also add the `PackageNodeExportTransport`?

src/interfaces/package-export-transport.ts
```diff
@@ -48,18 +48,16 @@ export interface PackageKeyAndVersionPair {
...
     key: string;
     parentNodeKey: string;
-    packageNodeKey: string;
     name: string;
     type: string;
     exportSerializationType: string;
     serializedContent: string;
```
ksalihu, line 54: Should we also rename `serializedContent` to `configuration`?

src/interfaces/package-export-transport.ts
promeris: The property is not used in CLI, and having it defined or not doesn't impact the parse/stringify steps during import.

src/interfaces/package-export-transport.ts
promeris: There is no benefit in doing this

Buqeta (dismissed):

Buqeta (approved):

ksalihu (approved):

jetakasabaqi (approved):

PR #184, opened 2024-05-29
TA-3012: Add diff packages command

Buqeta:
#### Description

Added a new command to diff the configuration of a sent zip of packages and their nodes with their configuration on production:

`config diff --profile (profile) --file (the zip of packages) --json --hasChanges `

This command has a few options:

- `--file` - the zip of packages that will be diffed. This file should come from `config export` command.
- `--profile` - the profile that references the team/realm where we want to get the production packages that will be diffed. It's a required option.
- `--hasChanges` - option that when sent on the command, sets only the `packageKey` and `hasChanges` flag for each package transport, meanwhile the `packageChanges` and `nodesWithChanges` arrays are left as empty.
- `--json` - when this option is sent on the command, the response is saved in a JSON file. By default, when this option is not sent, the response is written on the console as a prettified JSON.

#### Other

- Set `sourceMap` to true while running Jest tests, so that the code can be debugged while writing the tests.

#### Checklist

### DIFF
--- a/src/content-cli-config.ts
+++ b/src/content-cli-config.ts
@@ -70,6 +70,22 @@ export class Config {
 
         return program;
     }
+
+    public static diff(program: CommanderStatic): CommanderStatic {
+        program
+            .command("diff")
+            .description("Command to diff configs of packages")
+            .option("-p, --profile <profile>", "Profile of the team/realm which you want to use to diff the packages with")
+            .option("--hasChanges", "Flag to return only the information if the package has changes without the actual changes")
+            .option("--json", "Return the response as a JSON file")
+            .requiredOption("-f --file <file>", "Exported packages file (relative path)")
+            .action(async cmd => {
+                await new ConfigCommand().diffPackages(cmd.file, cmd.hasChanges, cmd.json);
+                process.exit();
+            });
+
+        return program;
+    }
 }
 
 const loadAllCommands = () => {
@@ -77,6 +93,7 @@ const loadAllCommands = () => {
     Config.listVariables(commander);
     Config.export(commander);
     Config.import(commander);
+    Config.diff(commander)
     commander.parse(process.argv);
 };
 
--- a/src/content-cli-config.ts
+++ b/src/content-cli-config.ts
@@ -78,7 +78,7 @@ export class Config {
             .option("-p, --profile <profile>", "Profile of the team/realm which you want to use to diff the packages with")
             .option("--hasChanges", "Flag to return only the information if the package has changes without the actual changes")
             .option("--json", "Return the response as a JSON file")
-            .requiredOption("-f --file <file>", "Exported packages file (relative path)")
+            .requiredOption("-f --file <file>", "Exported packages file (relative or absolute path)")
             .action(async cmd => {
                 await new ConfigCommand().diffPackages(cmd.file, cmd.hasChanges, cmd.json);
                 process.exit();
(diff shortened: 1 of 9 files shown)


tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "declarationDir": "./dist/lib",
     "noImplicitAny": false,
     "target": "es2015",
-    "sourceMap": false,
+    "sourceMap": true,
```
promeris, line 8: Why do we need to enable sourceMaps?

Buqeta: Yes, the `.map` files get generated for each `.js` file on `dist` folder. Why do you think sourceMaps should not be generated?

LaberionAjvazi (dismissed):

tsconfig.json
qmucolli: If we want to enable source maps, we cannot do so for productive builds as this exposes the code to be visible to all the users of content-cli which typically exposes us to the risk of reverse engineering. These rules do not apply as strictly for open source projects, however if we only need it for debugging, we should use sourceMaps only during the development builds and not productive ones.

promeris (dismissed):

tsconfig.json
Buqeta: @qmucolli I set `sourceMap` to true only when running Jest tests in [this commit](https://github.com/celonis/content-cli/pull/184/commits/139cb2bf090f8c3ef0f9a30cd2013fcb803af3bd). This works when debugging the code and doesn't generate source maps when running `yarn build`. Can you check it out?

LaberionAjvazi (dismissed):

src/content-cli-config.ts
```diff
@@ -70,13 +70,30 @@ export class Config {
...
     }
+
+    public static diff(program: CommanderStatic): CommanderStatic {
+        program
+            .command("diff")
+            .description("Command to diff configs of packages")
+            .option("-p, --profile <profile>", "Profile of the team/realm which you want to use to diff the packages with")
+            .option("--hasChanges", "Flag to return only the information if the package has changes without the actual changes")
+            .option("--json", "Return the response as a JSON file")
+            .requiredOption("-f --file <file>", "Exported packages file (relative path)")
```
ksalihu, line 81: I think this should work even if the passed file is absolute, right?

Buqeta: Yes. It works with both actually. I'll update it.

Buqeta: Changed it [here](https://github.com/celonis/content-cli/pull/184/commits/04bb9d7f9c9017295c517198c5d4f7c591823085).

src/services/package-manager/diff-service.ts
```diff
@@ -0,0 +1,73 @@
+import * as AmdZip from "adm-zip";
```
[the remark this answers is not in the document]
Buqeta: Nice catch! Updated it [here](https://github.com/celonis/content-cli/pull/184/commits/a56868f7c42449832f514e57181a07388f4cf3f8).

ksalihu (dismissed):

src/content-cli-config.ts
```diff
@@ -70,13 +70,30 @@ export class Config {
...
     }
+
+    public static diff(program: CommanderStatic): CommanderStatic {
+        program
+            .command("diff")
+            .description("Command to diff configs of packages")
+            .option("-p, --profile <profile>", "Profile of the team/realm which you want to use to diff the packages with")
+            .option("--hasChanges", "Flag to return only the information if the package has changes without the actual changes")
+            .option("--json", "Return the response as a JSON file")
+            .requiredOption("-f --file <file>", "Exported packages file (relative or absolute path)")
```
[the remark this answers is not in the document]
Buqeta: Nice catch! Added it [here](https://github.com/celonis/content-cli/pull/184/commits/d0d696db1a7b151fb090d42dbe4c852e56c5dae5).

ZgjimHaziri (approved):

ksalihu (approved):

LaberionAjvazi (approved):

promeris (approved):

PR #189, opened 2024-08-20
TA-3254: Update diff transports

Buqeta:
#### Description

Updated the diff transports to include the base package and nodes' configurations and updated the tests accordingly.

#### Checklist

### DIFF
--- a/src/interfaces/diff-package.transport.ts
+++ b/src/interfaces/diff-package.transport.ts
@@ -8,7 +8,7 @@ export interface ConfigurationChangeTransport {
 export interface NodeDiffTransport {
     nodeKey: string;
     changes: ConfigurationChangeTransport[];
-    baseConfiguration: NodeConfiguration
+    baseConfiguration: NodeConfiguration;
 }
 
 export interface NodeConfiguration {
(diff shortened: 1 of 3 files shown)


src/interfaces/diff-package.transport.ts
```diff
@@ -8,14 +8,38 @@ export interface ConfigurationChangeTransport {
 export interface NodeDiffTransport {
     nodeKey: string;
     changes: ConfigurationChangeTransport[];
+    baseConfiguration: NodeConfiguration
```
promeris, line 11: Can we end with semi-colon for all of these

LaberionAjvazi (dismissed):

src/interfaces/diff-package.transport.ts
Buqeta: Changed it [here](https://github.com/celonis/content-cli/pull/189/commits/7ab2d7015ee7831f6563e372852778267701f271).

jetakasabaqi (approved):

ksalihu (approved):

ZgjimHaziri (approved):

promeris (approved):

PR #190, opened 2024-08-20
[EMA-4088 & EMA-4345]: Integrate Action Flow Import/Export functionality

jetmirhalili11:
#### Description

This PR integrates the functionality of import/export of action flows

Three new commands introduced
- `analyze action-flows` - returns the action-flow metadata and its dependencies for a package
- `export action-flows` - returns the action-flow configuration and its dependencies for a package
- `import action-flows` - imports action-flows on a package

#### Checklist


Shpat-Jashari (dismissed):

src/content-cli-analyze.ts
```diff
@@ -0,0 +1,43 @@
...
+
+type CommanderStatic = commander.CommanderStatic;
+
+class Analyze {
+    public static actionFlows(program: CommanderStatic): CommanderStatic {
+        program
+            .command("action-flows")
+            .description("Analyze Action Flows dependencies for a certain package")
+            .option("-p, --profile <profile>", "Profile which you want to use to analyze Action Flows")
+            .requiredOption("-i, --packageId <packageId>", "ID of the package from which you want to export Action Flows")
```
promeris, line 14: I find the short hand notation very confusing, so would suggest to remove it.
```suggestion
            .requiredOption("--packageId <packageId>", "ID of the package from which you want to export Action Flows")
```

src/content-cli-analyze.ts
```diff
@@ -0,0 +1,43 @@
...
+            .option("-o, --outputToJsonFile", "Output the analyze result in a JSON file")
```
promeris, line 15: Let's name this option `--json` to stay consistent with other commands.
```suggestion
            .option("--json", "Return response as json type", "")
```

Edit: Please apply the comments for all the commands.

src/services/action-flow/action-flow-service.ts
```diff
@@ -0,0 +1,72 @@
...
+        formData.append("file", fs.createReadStream(fileName, { encoding: null }), { filename: fileName });
+
+        return formData;
+    }
+
+    private attachMetadataFile(fileName: string, zip: AdmZip): void {
+        fileName = fileName + (fileName.endsWith(".json") ? "" : ".json");
+        const metadata = fileService.readFile(fileName);
+
+        zip.addFile("metadata.json", Buffer.from(metadata));
```
promeris, line 68: Let's make this `metadata.json` a constant in the file

promeris (commented):
Can you also include unit tests for the changes? You can follow some examples we have already, and you can reach out to @LaberionAjvazi if you need any further support.

src/content-cli-analyze.ts
jetmirhalili11: I tried to follow the other commands in `import` and `export`
- export data-pool - https://github.com/celonis/content-cli/blob/141ad700180f8f2a89dad80ac35b0baac312dc90/src/content-cli-export.ts#L31
- import data-pools - https://github.com/celonis/content-cli/blob/141ad700180f8f2a89dad80ac35b0baac312dc90/src/content-cli-import.ts#L36

I can see that `--json` is used at `config` and `list` commands, but I'm not sure if we should be consistent with that, or with the other commands in the same file

src/services/action-flow/action-flow-service.ts
jetmirhalili11: Handled at [179f174](https://github.com/celonis/content-cli/pull/190/commits/179f174f5f59fbcf2817febdf0311a02e2974229)

src/content-cli-analyze.ts
jetmirhalili11: Handled at [179f174](https://github.com/celonis/content-cli/pull/190/commits/179f174f5f59fbcf2817febdf0311a02e2974229)

AlbinHoxha (dismissed):

DOCUMENTATION.md
```diff
@@ -440,6 +444,59 @@ the same command as with pushing other assets to Studio:
...
+
+#### Import Action Flows
+
+The import operation allows import of Action Flows of one package together with their dependencies. Dependencies
+could be webhooks, data structures, variables and other.
+
+In order to push Action Flows use the following command:
+
+```
+content-cli import action-flows -p my-profile-name -i <replace-with-package-id> -f <replace-with-exported-zip-file> --dryRun --outputToJsonFile
```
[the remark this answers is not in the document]
jetmirhalili11: Handled at [5a19c77](https://github.com/celonis/content-cli/pull/190/commits/5a19c77a2184b6a41731f0898fe80bad0259706d)

src/content-cli-analyze.ts
promeris: hmmm, I guess we already have this format for this command. Let's leave like this for now, we can look into unifying the options in the future.

promeris (commented):
One more note: To publish the changes you need to manually bump the version in `package.json`. Would suggest bumping only patch version, until we decide to release properly with documentation

DOCUMENTATION.md
```diff
@@ -440,6 +444,59 @@ the same command as with pushing other assets to Studio:
...
+In order to push Action Flows use the following command:
+
+```
+content-cli import action-flows -p my-profile-name --packageId <replace-with-package-id> -f <replace-with-exported-zip-file> --dryRun --outputToJsonFile
+```
+
+##### Input
+
+* The zip file is the one that you receive from the ```action-flows export``` command
+* The ```--outputToJsonFile``` is optional. If specified, the import will be executed on dry run mode (nothing will be created)
```
jetakasabaqi, line 491: This description should be for --dryRun, not --outputToJsonFile right?

src/services/action-flow/action-flow-service.ts
```diff
@@ -0,0 +1,74 @@
...
+
+class ActionFlowService {
+    private static readonly METADATA_FILE_NAME = "metadata.json";
+
+    public async exportActionFlows(packageId: string, metadataFilePath: string): Promise<void> {
+        const exportedActionFlowsData = await actionFlowApi.exportRawAssets(packageId);
+        const tmpZip: AdmZip = new AdmZip(exportedActionFlowsData);
+
+        const zip = new AdmZip();
+        tmpZip.getEntries().forEach(entry => {
```
jetakasabaqi, line 17: Why are we creating a new zip? Can you not use the tmpZip?

DOCUMENTATION.md
jetmirhalili11: Yes nice catch
however at [54164d6](https://github.com/celonis/content-cli/pull/190/commits/54164d610690d4149f1e123f35b3e52943328e54) I removed the import commands from the documentation (including these likes)

src/services/action-flow/action-flow-service.ts
jetmirhalili11: Somehow adding a zip with entries using the `adm-zip` constructor differs from adding the zip entries using the `addFile` method

If I used the constructor to create the zip, and then send that zip to the server I was receiving the error `invalid entry size (expected 2187395080 but got 158 bytes)` therefore I had to use this kind of a hacky workaround.
I also spent some time looking at the code of `AdmZip` on why this is happening and noticed that they have a flag in case the object is created with another zip on the constructor and they use it in a lot of places, but it was hard to find something else or something more concrete.
However, the bug was for sure in the library as also creating the zip manually or with other tools worked as expected

Btw I tried to explain this issue a little bit also on the [commit message](https://github.com/celonis/content-cli/pull/190/commits/4a60c8bcb78beff4d4ac9a3b5cf82e345a97acdb)

promeris (dismissed):

src/services/action-flow/action-flow-service.ts
jetakasabaqi: Thanks for explaining!

jetakasabaqi (dismissed):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@celonis/content-cli",
-  "version": "0.10.0",
+  "version": "0.10.1",
```
Buqeta, line 3: Can you please increase the minor version to 11, because I just merged [a PR](https://github.com/celonis/content-cli/pull/189) that sets the version to `0.11.0`?

LaberionAjvazi (dismissed):

Buqeta (approved):

promeris (approved):

LaberionAjvazi (approved):

PR #191, opened 2025-01-23
TN-6665: Upgrade to latest version of dependencies to resolve security vulnerabilities

qmucolli:
#### Description

This PR contains the changes necessary to resolve the security vulnerabilities found by our SAST scans, some libraries were removed altogether (like vm2), and in some cases only the usages of the library had to be updated (e.g. commander).

#### Checklist

  - Verified that every command listed in the content-cli --help was still working as it used to.

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Setup Node with Github Registry
         uses: actions/setup-node@v1
         with:
-          node-version: '16'
+          node-version: '20'
           registry-url: https://npm.pkg.github.com/celonis
           scope: '@celonis'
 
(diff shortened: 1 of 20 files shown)


.github/workflows/build-and-publish.yml
```diff
@@ -56,8 +56,3 @@ jobs:
         name: Checkout Master
         with:
           fetch-depth: '0'
-
-      - uses: tool3/bump@v1
-        name: Bump patch version
-        with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
```
diellzacunaj, line 63: nit: I'd say it's best to make bump the minor version automatically

diellzacunaj (approved):

.github/workflows/build-and-publish.yml
qmucolli: It seems that its not working anymore and eitherway we have to come back to this, so for now I've removed this so we can think of a better solution.

PR #194, opened 2025-03-12
TA-3622: Downgrade AdmZip version

promeris:
#### Description

Downgrade AdmZip library to a version where zip descriptors are not checked. The nested zips we are generating during `config export/import` commands create zips that don't contain descriptors, and the library we are using doesn't provide with any way to force zip descriptors to be written during exports. Issues were introduced with this [PR](https://github.com/cthackers/adm-zip/pull/508)

Downgrade will offer a temporary solution until either a fix is provided in the library itself, or otherwise we will need to switch libraries.

Ticket: https://celonis.atlassian.net/browse/TA-3622

#### Checklist

  - Tested `config export` and `config import`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celonis/content-cli",
-  "version": "0.13.2",
+  "version": "0.13.3",
   "description": "CLI Tool to help manage content in Celonis EMS",
   "main": "content-cli.js",
   "bin": {
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   },
   "dependencies": {
     "@datadog/datadog-ci": "2.48.0",
-    "adm-zip": "0.5.16",
+    "adm-zip": "0.5.14",
     "axios": "1.7.9",
     "commander": "13.1.0",
     "form-data": "4.0.1",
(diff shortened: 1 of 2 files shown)


jetakasabaqi (dismissed):

package.json
```diff
@@ -19,7 +19,7 @@
   },
   "dependencies": {
     "@datadog/datadog-ci": "2.48.0",
-    "adm-zip": "0.5.16",
+    "adm-zip": "0.5.14",
```
jetakasabaqi, line 22: Do you need to increase the content-cli version also?

ksalihu (approved):

Buqeta (approved):

ZgjimHaziri (approved):

jetakasabaqi (approved):

qmucolli:
I'm curious to understand how the extraction of keys happens in the team to team copy as I've also tested this flow locally and it seemed to be working fine.

On a side-note, there was a vulnerability reported with versions older than the latest version of adm-zip, can we make sure to run the SAST scan once this PR is merged to verify that we're not introducing new vulnerabilities?

promeris:
> I'm curious to understand how the extraction/import of zip happens in the team to team copy as I've also tested this flow locally and it seemed to be working fine.
> 
> On a side-note, there was a vulnerability reported with versions older than the latest version of adm-zip, can we make sure to run the SAST scan once this PR is merged to verify that we're not introducing new vulnerabilities?

Team to team copy is using the `content-cli config export/import` commands, and when testing through UI it's using a fixed version now and not auto-updating, so not a breaking change immediately, however when testing through only content-cli, you can bump into the issue with no present descriptor.

I can check the SAST scan after merging, since the actions currently don't provide with option to run a branch.

qmucolli (approved):

PR #199, opened 2025-05-14
TA-3750: Setup command modules

ZgjimHaziri:
#### Description

Defined the needed command module directories and added the base `module.ts` file. Also added how will the codeowners look after the changes. Left them commented for now, and will uncomment as part of the tickets that migrate the commands of each area.

#### Checklist


LaberionAjvazi (dismissed):

Buqeta (dismissed):

jetakasabaqi (dismissed):

ksalihu (dismissed):

.github/CODEOWNERS
```diff
@@ -1 +1,12 @@
 *   @celonis/navi @gencblakqorip9 @celonis/astro
+
+#/*                              @celonis/astro
+#/src                            @celonis/astro
+#/src/core                       @celonis/astro
+#/src/commands/pacman/           @celonis/astro
+#/src/commands/profile/          @celonis/astro
```
promeris, line 7: The first line `*` should cover these paths as well. 
```suggestion
```

promeris (dismissed):

.github/CODEOWNERS
ZgjimHaziri: The `*` would define the set owners everywhere in the repo. We would not like to have ownership of the other team commands. Similar to cfg-ibc ownership.

promeris: So if we forget to add codeowners for a new folder structure, we would rather have it orphan and not assign a fallback owner?

ZgjimHaziri: Do you have another idea in mind?

promeris: I was thinking that we should leave `* @celonis/astro`, and only specify overrides of ownership for specific folder structures. This would make sure all changes are covered from repository rules and require PRs for respective owners.

ZgjimHaziri: Astro would get added as an unnecessary reviewer for each command change. I don't think we would want that.

Either way, we will have the `* @celonis/astro` until the refactoring is done, so we can decide then if we want to remove it.

ZgjimHaziri:
Updates:
- Renamed `pacman` module to `config`. Because Pacman is not a good description to features provided there.
- After discussing with @meritonhusaj and @GencBlakqoriP9, removed non-Astro codeowners during the refactoring.

ksalihu:
> * Renamed `pacman` module to `config`. Because Pacman is not a good description to features provided there.

Are we making an exception to the owning area naming convention, or are we considering `config` to be the owning area?

ZgjimHaziri:
> Are we making an exception to the owning area naming convention, or are we considering `config` to be the owning area?

Agree. Changed it to `configuration-management`, because it's more of an area.

ksalihu (dismissed):

.github/CODEOWNERS
```diff
@@ -1 +1,12 @@
-*   @celonis/navi @gencblakqorip9 @celonis/astro
+*                               @celonis/astro
+
+#/*                              @celonis/astro
+#/src                            @celonis/astro
+#/src/core                       @celonis/astro
+#/src/commands/pacman/           @celonis/astro
```
jetakasabaqi, line 6: this should say configuration-managment now

ZgjimHaziri: Updated.

Buqeta (approved):

promeris (approved):

jetakasabaqi (approved):

PR #215, opened 2025-06-16
TA-3752: Deprecate Studio export/import commands

ZgjimHaziri:
#### Description

- Added deprecation notice for batch export/import commands in Studio.
- The deprecation notice has been set to ~2 months later (1st of September)
- The notices points to the `config export/import` commands.
- Deprecation done based on this [sheet](https://docs.google.com/spreadsheets/d/1aZ5D4_ss412lI8jAoOZstqAen2Od8pBwlAPpCmhgYW8/edit?gid=0#gid=0).

Example of a deprecation message:
```
warn:    ⚠️  [DEPRECATION NOTICE] 
`content-cli import packages` is deprecated and is expected to be removed in subsequent updates around: 01-09-2025.
Please use `content-cli config import` instead.
```

Implemented a built-in deprecation notice printer which can be reused. 

#### Checklist

### DIFF
--- a/src/commands/studio/module.ts
+++ b/src/commands/studio/module.ts
@@ -19,6 +19,8 @@ class Module extends IModule {
             .requiredOption("--packageKeys <packageKeys...>", "Exports only given package keys")
             .option("--includeDependencies", "Include variables and dependencies", "")
             .option("--excludeActionFlows", "Don't export action flows")
+            .deprecationNotice("`content-cli export packages` is deprecated and is expected to be removed in subsequent updates around: 01-09-2025.\n" +
+                "Please use `content-cli config export` instead.\n")
             .action(this.batchExportPackages);
 
         const importCommand = configurator.command("import");
@@ -32,6 +34,8 @@ class Module extends IModule {
             .option("--excludeActionFlows", "Skip overwrite of action flows of package")
             .option("--dataModelMappingsFile <dataModelMappingsFile>", "DataModel variable mappings file path. If missing, variables will be mapped from manifest file.")
             .requiredOption("-f, --file <file>", "Exported packages file (relative path)")
+            .deprecationNotice("`content-cli import packages` is deprecated and is expected to be removed in subsequent updates around: 01-09-2025.\n" +
+                "Please use `content-cli config import` instead.\n")
             .action(this.batchImportPackages);
 
         const listCommand = configurator.command("list");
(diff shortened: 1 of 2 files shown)


fatimarmullaku (approved):

src/commands/studio/module.ts
```diff
@@ -19,6 +19,8 @@ class Module extends IModule {
             .requiredOption("--packageKeys <packageKeys...>", "Exports only given package keys")
             .option("--includeDependencies", "Include variables and dependencies", "")
             .option("--excludeActionFlows", "Don't export action flows")
+            .deprecationNotice("`content-cli export packages` is deprecated and is expected to be removed in subsequent updates around: 01-09-2025.\n" +
```
Buqeta, line 22: If we're providing these new commands to the users, I think we should first create an official documentation for them in the repo, since right now these commands are "hidden" in that sense.

ksalihu (approved):

LaberionAjvazi (approved):

src/commands/studio/module.ts
ZgjimHaziri: Looked into it and considering that this command doesn't really have much usage, would propose to leave it like this until we have a clear plan on how will we proceed with the documentation of those commands. Will see if others have an opinion on it.

Buqeta: Which command doesn't have much usage? The Studio export? If that's what you meant, how do we know if it doesn't?

jetakasabaqi (approved):

src/commands/studio/module.ts
Buqeta: Thanks! Then I guess it's fine to deprecate them.

Buqeta (approved):

src/commands/studio/module.ts
ZgjimHaziri: Extra note: the `axios` user-agent calls seem to be coming from content-cli too, but from the old `http-client.service.ts`. The export/import commands of Studio use the newer `HttpClientServiceV2` which always appends the user-agent as content-cli. That's extra assurance on them not being used by these commands but only by the single export/import ones.

Buqeta: That's great! Thanks for checking! 🎉

fisgeci (approved):

PR #228, opened 2025-07-04
TA-3920: Fix global install of Content CLI

ZgjimHaziri:
#### Description

When installing globally, `npm` attempts to create a link to the `./content-cli.js` of the project. This fails since the actual js file is located inside of the `dist` directory.

**Additional details**
The dist was removed in order to make it easier to do local `npm link` without cd-ing in the dist. Added this change in the local setup guide too.

#### Checklist

  - Tested the npm link command. Testing the global install can be done only after publishing.  

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@celonis/content-cli",
   "version": "1.0.1",
-  "description": "CLI Tool to help manage content in Celonis EMS",
+  "description": "CLI Tool to help manage content in Celonis Platform",
   "main": "content-cli.js",
   "bin": {
     "content-cli": "./content-cli.js"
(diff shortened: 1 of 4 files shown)


Buqeta (approved):

LaberionAjvazi (approved):

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "CLI Tool to help manage content in Celonis EMS",
   "main": "content-cli.js",
   "bin": {
-    "content-cli": "./dist/content-cli.js"
+    "content-cli": "./content-cli.js"
```
ksalihu, line 7: I don't understand, from the PR description

> the actual js file is located inside of the dist directory.

doesn't this mean that the file should be referenced via the `dist` path as it is right now?

ksalihu (approved):

Buqeta (approved):

PR #230, opened 2025-07-04
TA-3902: Include Package ID in Studio's `list packages` command response

ZgjimHaziri:
#### Description

The `list packages` command in Studio doesn't return the Package ID, when being called without the `includeDependencies` flag. This decision seems to have been made because of the fact that the ID was never used in the cases when it was called without the flag.

In a recent case (detailed in the linked ticket), we're using the command in order to get the ID, which gets used for the `action-flows` import. Proposing to add the ID field in the response file by default, not conditional on the `includeDependencies` parameter.

**Note**
The version bump is being done in the previous major version.

#### Checklist

  - Tested the response for both staging and published packages. All included correct IDs (as verified from the frontend response of those Packages). 

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celonis/content-cli",
-  "version": "0.13.5",
+  "version": "0.13.6",
   "description": "CLI Tool to help manage content in Celonis EMS",
   "main": "content-cli.js",
   "bin": {
--- a/src/services/package-manager/package-service.ts
+++ b/src/services/package-manager/package-service.ts
@@ -35,7 +35,7 @@ class PackageService {
     }
 
     public async findAndExportListOfAllPackages(includeDependencies: boolean, packageKeys: string[]): Promise<void> {
-        const fieldsToInclude = ["key", "name", "changeDate", "activatedDraftId", "workingDraftId", "spaceId"];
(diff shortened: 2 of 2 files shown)


LaberionAjvazi (approved):

Buqeta (approved):

Buqeta (approved):

ksalihu (approved):

jetakasabaqi (approved):

promeris (approved):

fisgeci (approved):

gramozkaragjyzi (changes requested):
Can we please evaluate if we really need to do this? As we want to move away from ids, really.

ZgjimHaziri:
> Can we please evaluate if we really need to do this? As we want to move away from ids, really.

The way I see it, we're not adding any ID usage with the implementation because:
- This command is already returning the ID, just not in this case. The effort for removing ID usage from here in the future would remain the same.
- The usage of this command from CDP, is already utilizing IDs regardless of what command is used. That's because of the Action Flows commands that work with Package ID.

Agree that the right way to approach this is for the Action Flows to switch to using keys, but as a temporary fix, this doesn't seem to add complexity in our way towards using Keys. The fix is being done to unblock client usage and development. 

Let me know what you think.

gramozkaragjyzi (approved):

PR #232, opened 2025-07-07
TA-3920: Fix package.json path

ZgjimHaziri:
#### Description

The VersionUtils which logs the version of the CLI in the console, has the path of the required `package.json` wrong when trying to install it globally. The reason this is working locally is because of the fact that the directory locally is structured that way that when you go up in the tree from `dist`, you will also find the original `package.json` there. 

#### Special notes

Unpublished the 1.0.1 version that was published with a faulty path.

#### Checklist

### DIFF
--- a/src/core/utils/version.ts
+++ b/src/core/utils/version.ts
@@ -1,5 +1,5 @@
 // tslint:disable-next-line:no-var-requires
-const { version } = require("../../../package.json");
+const { version } = require("../../package.json");
 
 export class VersionUtils {
     public static getCurrentCliVersion(): string {


LaberionAjvazi (dismissed):

promeris (approved):

Buqeta (dismissed):

jetakasabaqi (dismissed):

ZgjimHaziri:
The initial solution of this PR was working for both global and local installations, but was failing the tests in which the location of `package.json` is pretty different. As a solution added the [package-up](https://www.npmjs.com/package/package-up) package, which is a simple tool for finding the closest `package.json`. 

Note: since the tool does async file reading, had to do some minor refactorings in the `content-cli.ts` file to fit the async changes.

promeris (approved):

Buqeta (approved):

ksalihu:
> The initial solution of this PR was working for both global and local installations, but was failing the tests in which the location of `package.json` is pretty different. As a solution added the [package-up](https://www.npmjs.com/package/package-up) package, which is a simple tool for finding the closest `package.json`.
> 
> Note: since the tool does async file reading, had to do some minor refactorings in the `content-cli.ts` file to fit the async changes.

Checked the library a bit, the implementation is kinda simple, I don't think we should use a library for these checks.
In the tests maybe we should see if we can mock the version util or something like that

ksalihu (approved):

PR #234, opened 2025-07-11
TA-3947: Fix the issues with importing version

promeris:
#### Description

- Reverted changes of adding `package-up` package.
- Implemented different approach to fix the test when the `package.json` is imported
  - Changed the path of the mock to match the new path used in `version.ts`

#### Checklist


ZgjimHaziri (approved):

LaberionAjvazi (approved):

ksalihu:
What was the problem?

promeris:
> What was the problem?

@ksalihu Pretty much no command was working after installing this version, due to this library not being able to find the `package.json`

PR #239, opened 2025-07-18
TA-3971: Add option to export packages by versions

Buqeta:
#### Description

- Added an option to the existing `config export` command `--keysByVersion` which exports the packages by the specific received versions.
- Added logic to check if both versions are being sent or neither of them isn't sent, so the command fails with a message.
- Used the same batch export logic, just added a check that controls which batch export endpoint will be called based on the received option.
- Added test cases that are the same as batch export cases.
- Updated CLI version.

#### Checklist

### DIFF
--- a/src/commands/configuration-management/api/batch-import-export-api.ts
+++ b/src/commands/configuration-management/api/batch-import-export-api.ts
@@ -62,9 +62,9 @@ export class BatchImportExportApi {
         });
     }
 
-    public async exportPackagesByVersions(packageKeysByVersion: string[], withDependencies: boolean = false): Promise<Buffer> {
+    public async exportPackagesByVersions(packageKeysWithVersion: string[], withDependencies: boolean = false): Promise<Buffer> {
         const queryParams = new URLSearchParams();
-        packageKeysByVersion.forEach(packageKeyByVersion => queryParams.append("packageKeysWithVersion", packageKeyByVersion));
+        packageKeysWithVersion.forEach(packageKeyByVersion => queryParams.append("packageKeysWithVersion", packageKeyByVersion));
         queryParams.set("withDependencies", withDependencies.toString());
 
         return this.httpClient().getFile(`/package-manager/api/core/packages/versions/export/batch?${queryParams.toString()}`).catch(e => {
--- a/src/commands/configuration-management/api/batch-import-export-api.ts
+++ b/src/commands/configuration-management/api/batch-import-export-api.ts
@@ -62,6 +62,16 @@ export class BatchImportExportApi {
         });
     }
 
+    public async exportPackagesByVersions(packageKeysByVersion: string[], withDependencies: boolean = false): Promise<Buffer> {
+        const queryParams = new URLSearchParams();
+        packageKeysByVersion.forEach(packageKeyByVersion => queryParams.append("packageKeysWithVersion", packageKeyByVersion));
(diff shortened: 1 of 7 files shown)


src/commands/configuration-management/api/batch-import-export-api.ts
```diff
@@ -62,6 +62,16 @@ export class BatchImportExportApi {
         });
     }
 
+    public async exportPackagesByVersions(packageKeysByVersion: string[], withDependencies: boolean = false): Promise<Buffer> {
```
jetakasabaqi, line 65:
```suggestion
    public async exportPackagesByVersions(packageKeysWithVersion: string[], withDependencies: boolean = false): Promise<Buffer> {
```

src/commands/configuration-management/module.ts
```diff
@@ -24,9 +24,15 @@ class Module extends IModule {
 
         configCommand.command("export")
             .description("Command to export package configs")
-            .requiredOption("--packageKeys <packageKeys...>", "Keys of packages to export")
+            .option("--packageKeys <packageKeys...>", "Keys of packages to export. Exports the latest deployed version only")
+            .option("--keysByVersion <keysByVersion...>", "Keys of packages to export by version")
```
jetakasabaqi, line 28: Can we name this packageKeysWithVersion to match the API?

ksalihu (commented):
I think the command should be `config versions export` `{resource} {action}`

src/commands/configuration-management/api/batch-import-export-api.ts
Buqeta: Changed it [here](https://github.com/celonis/content-cli/pull/239/commits/77a607238b99bc22e8356dbc6d75eec111763a0c).

Buqeta:
> I think the command should be `config versions export` `{resource} {action}`

We're technically exporting packages. The version is just a property of the package. I don't think it's a resource on its own. This structure you're proposing is also confusing to the CLI user since the `config export` exports packages and it doesn't have the `packages` name in the command.

src/commands/configuration-management/module.ts
```diff
@@ -24,9 +24,15 @@ class Module extends IModule {
...
             .option("--withDependencies", "Include variables and dependencies", "")
-            .action(this.batchExportPackages);
+            .action(async (context: Context, command: Command, options: OptionValues): Promise<void> => {
+                if ((options.packageKeys && options.keysByVersion) || (!options.packageKeys && !options.keysByVersion)) {
+                    throw new Error("Please provide either --packageKeys or --keysByVersion, but not both.");
+                }
```
ZgjimHaziri, line 33: I think it's better if we keep the action callback clean here, and implement the validation on the `batchExportPackages` method.

Buqeta: Changed it [here](https://github.com/celonis/content-cli/pull/239/commits/f04b5b6c60be7cc2af7129ba4cf6061358e7459a).

ZgjimHaziri (approved):

jetakasabaqi (approved):

ZgjimHaziri:
> We're technically exporting packages. The version is just a property of the package. I don't think it's a resource on its own. This structure you're proposing is also confusing to the CLI user since the `config export` exports packages and it doesn't have the `packages` name in the command.

The main point here is that the existing export command is not in a proper structure.

PR #244, opened 2025-08-12
TA-4103: Zip/Unzip in export/import

ZgjimHaziri:
#### Description

Enable exporting files as unzipped and importing unzipped directories.
- Added a `--unzip` option in export which decides if the result of the download is a zip or a directory.
- In import we added a `--directory` option as an alternative to `--file`.
  - Added validation for the different combinations of the import options.

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celonis/content-cli",
-  "version": "1.1.1",
+  "version": "1.1.2",
   "description": "CLI Tool to help manage content in Celonis Platform",
   "main": "content-cli.js",
   "bin": {


Buqeta (approved):

promeris (approved):

src/commands/configuration-management/batch-import-export.service.ts
```diff
@@ -121,11 +119,12 @@ export class BatchImportExportService {
         let fileToBeImported: string;
         if (gitBranch) {
             fileToBeImported = await this.gitService.pullFromBranch(gitBranch);
-            fileToBeImported = fileService.zipDirectoryInBatchExportFormat(fileToBeImported);
         } else {
             fileToBeImported = file;
         }
 
+        fileToBeImported = fileService.zipDirectoryInBatchExportFormat(fileToBeImported);
```
jetakasabaqi, line 126: What if the file is already a zip?

jetakasabaqi (approved):

ksalihu (approved):

PR #247, opened 2025-08-21
TA-4128: Handle client errors in downloadFile

ksalihu:
#### Description

Handled the response manually for different scenarios.
- When the response is 200, then the response type is `application/octet-stream` which requires handling differently, but in the case of 404 or 403, the response was json or plain text so the data was being treated as if its a file and getting zipped, hence why the response was unreadable.
- Now the data handling is done manually

#### Checklist

### Changes

Before:
image
Note: the full output is way longer than the screenshot

After:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@celonis/content-cli",
-  "version": "1.1.2",
+  "version": "1.1.3",
   "description": "CLI Tool to help manage content in Celonis Platform",
   "main": "content-cli.js",
   "bin": {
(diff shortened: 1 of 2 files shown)


Buqeta (approved):

ZgjimHaziri (approved):

ZgjimHaziri:
Let's not forget to bump the version as part of the PR.

ZgjimHaziri (approved):

promeris (approved):

LaberionAjvazi (approved):

PR #257, opened 2025-09-15
TA-4292: Remove token storing option in git profiles

ZgjimHaziri:
#### Description

- Removed token storing in git profiles, in favor of delegating the authentication to the local Git configurations. Now only the authentication method is selected, which decides which local setup will it use.
- Fixed environment variable authentication which will provide a way for token authentication in pipelines. This implementation was previously partial and undocumented.
- Added documentation for env var auth and modified the existing auth docs to fit the changes.

#### Relevant links

- Jira: https://celonis.atlassian.net/browse/TA-4292

#### Checklist

  - Will add automated tests now


Buqeta (commented):
Can we increase the CLI version?

promeris (approved):

PR #266, opened 2025-10-06
TA-4307: Make packageKey option required in the getTargets command.

jetakasabaqi:
#### Description

Made packageKey option requires in the getTargets commands to match the API.

#### Relevant links

- Jira: https://celonis.atlassian.net/browse/TA-4307

#### Checklist

### DIFF
--- a/src/commands/deployment/deployment-api.ts
+++ b/src/commands/deployment/deployment-api.ts
@@ -80,13 +80,10 @@ export class DeploymentApi {
             });
     }
 
-    public async getTargets(deployableType: string, packageKey?: string): Promise<TargetTransport[]> {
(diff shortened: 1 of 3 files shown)


src/commands/deployment/deployment-api.ts
```diff
@@ -80,13 +80,10 @@ export class DeploymentApi {
             });
     }
 
-    public async getTargets(deployableType: string, packageKey?: string): Promise<TargetTransport[]> {
+    public async getTargets(deployableType: string, packageKey: string): Promise<TargetTransport[]> {
```
promeris, line 83: Do we also need to make it a `requiredParam` on the command?

Buqeta (approved):

PR #303, opened 2026-01-13
TA-4814: Remove deprecated Studio batch export/import commands

ZgjimHaziri:
#### Description

Studio batch export import commands have been marked for deprecation and up for removal since 1 September 2025. Removing them now.

Deleted also unused code related to its removal.

#### Relevant links

- Jira: https://celonis.atlassian.net/browse/TA-4814

#### Checklist

  - Since there are no existing tests for these commands, we will not be adding additional tests for the services affected as part of the cleanup 


promeris (approved):

Buqeta (approved):
This probably requires a major version change, right?

ZgjimHaziri:
> This probably requires a major version change, right?

I would consider it a minor one, considering that the commands have been marked as deprecated for half a year now, and the reason for deprecation itself was the fact that these commands weren't being used.

jetakasabaqi (approved):

ksalihu (approved):

PR #335, opened 2026-04-13
TA-4732: restrict file permissions created by content-cli

kuvia:
#### Description

Prevent world writeable/readable files created by content-cli. Directories are created with user rwx permissions and files are created with user rw permissions only.

Note: Downloaded package zips that are not created by content-cli itself but that come from the API have more lenient permissions so when unpacked manually they will still have group/world permissions.

#### Relevant links

https://celonis.atlassian.net/browse/TA-4732

#### Checklist

### DIFF
--- a/src/commands/action-flows/action-flow/action-flow.service.ts
+++ b/src/commands/action-flows/action-flow/action-flow.service.ts
@@ -6,6 +6,7 @@ import { Context } from "../../../core/command/cli-context";
 import { ActionFlowApi } from "./action-flow-api";
 import { fileService, FileService } from "../../../core/utils/file-service";
 import { logger } from "../../../core/utils/logger";
+import { FileConstants } from "../../../core/utils/file.constants";
 
 export class ActionFlowService {
     public static readonly METADATA_FILE_NAME = "metadata.json";
@@ -22,7 +23,7 @@ export class ActionFlowService {
 
         const zip = new AdmZip();
         tmpZip.getEntries().forEach(entry => {
-            zip.addFile(entry.entryName, entry.getData(), "", 0o600);
+            zip.addFile(entry.entryName, entry.getData(), "", FileConstants.DEFAULT_FILE_PERMISSIONS);
         });
 
         if (metadataFilePath) {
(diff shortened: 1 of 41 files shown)


LaberionAjvazi (dismissed):

src/commands/action-flows/action-flow/action-flow.service.ts
```diff
@@ -22,15 +22,15 @@ export class ActionFlowService {
...
+            zip.addFile(entry.entryName, entry.getData(), "", 0o600);
         });
 
         if (metadataFilePath) {
             this.attachMetadataFile(metadataFilePath, zip);
         }
 
         const fileName = "action-flows_export_" + uuidv4() + ".zip";
-        zip.writeZip(fileName);
+        zip.writeZip(fileName, () => fs.chmodSync(fileName, 0o600));
```
promeris: can we save `const PRIVATE_FILE_MODE: number = 0o600;` somewhere and use it across files? I would much rather read PRIVATE_FILE_MODE than try to understand what the octal integer representation means

src/core/git-profile/git/git.service.ts
```diff
@@ -20,7 +20,7 @@ export class GitService {
         const repoUrl = this.getRepoUrl();
 
         const targetDir = path.join(os.tmpdir(), `content-cli-${uuid()}`);
-        fs.mkdirSync(targetDir, { recursive: true });
+        fs.mkdirSync(targetDir, { recursive: true, mode: 0o700 });
```
promeris: same here, something like `{ mode: DIRECTORY_MODE }` would be more readable, where `const DIRECTORY_MODE: number = 0o700;` is a documented constant

src/commands/action-flows/action-flow/action-flow.service.ts
kuvia: makes sense, I've added a new constants file. I left the assertions using octals though so any change to the default permissions should be a conscious decision.

promeris (dismissed):

ZgjimHaziri (dismissed):

vullnet-kurti (dismissed):
haven't reviewed the code; approving to unblock the merge which requires approval from `process-automation`

vullnet-kurti (approved):
https://github.com/celonis/content-cli/pull/335#pullrequestreview-4129064900

PR #341, opened 2026-04-16
SP-45: Add asset-registry schema, examples, methodology commands

ZgjimHaziri:
#### Description

Add dedicated `asset-registry` subcommands for fetching schemas, examples, and methodology — replacing the beta `api request` command that required manual URL construction.

New commands:
- `asset-registry schema --assetType X [--json]`
- `asset-registry examples --assetType X [--json]`
- `asset-registry methodology --assetType X [--json]`

These hit the new Pacman proxy endpoints added in [celonis/pacman#1646](https://github.com/celonis/pacman/pull/1646), so callers no longer need to resolve descriptor paths themselves.

The `api` beta command is removed — it's fully superseded by these commands.

The agentic development guide, Cursor skill, and command docs are updated accordingly.

#### Relevant links

- Jira: [SP-45](https://celonis.atlassian.net/browse/SP-45)
- Pacman proxy APIs: [celonis/pacman#1646](https://github.com/celonis/pacman/pull/1646)

#### Checklist

- [x] I have self-reviewed this PR
- [x] I have tested the change and proved that it works in different scenarios
- [x] I have updated docs if needed

#### AI
Includes-AI-Code: true

Made with [Cursor](https://cursor.com)

[SP-45]: https://celonis.atlassian.net/browse/SP-45?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #342, opened 2026-04-17
SP-45: Add asset-registry validate command

ZgjimHaziri:
## Summary

Add `asset-registry validate` command with three mutually exclusive modes for validating asset configurations:
- `--configuration`/`-c` with `--packageKey`: validate a configuration JSON before import (CLI builds the ValidateRequest envelope)
- `--nodeKey` with `--packageKey`: validate an already-stored node on the platform
- `-f`: provide a file with the full ValidateRequest body (standalone alternative to all other options)

Proxies to Pacman's `POST /core/asset-registry/validate/{assetType}` endpoint.

**Ticket**: [SP-45](https://celonis.atlassian.net/browse/SP-45)
**Depends on**: #341

## Test plan

- [x] 18 unit tests covering all three modes, mutual exclusivity errors, JSON parse errors, and missing required options
- [x] All 36 asset-registry tests pass locally
- [x] `content-cli asset-registry validate --help` output verified

[SP-45]: https://celonis.atlassian.net/browse/SP-45?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ksalihu (dismissed):

docs/user-guide/asset-registry-commands.md
```diff
@@ -71,6 +71,51 @@ Options:
...
+content-cli asset-registry validate --assetType BOARD_V2 -f request.json
+```
+
+### Options
+
+- `--assetType <assetType>` (required) – The asset type identifier
+- `--packageKey <packageKey>` – Package key containing the node
+- `--nodeKey <nodeKey>` – Key of an already-stored node to validate on the platform
+- `--configuration <configuration>` – Inline JSON of the configuration to validate before import
+- `-c, --configFile <configFile>` – Path to a JSON file containing the configuration to validate before import
```
promeris: the short hand looks very confusing considering we have both --configuration and --configFile. I would just remove it here and leave no short hand

docs/user-guide/asset-registry-commands.md
```diff
@@ -71,6 +71,51 @@ Options:
...
+
+Validate asset configurations against the asset service's validation endpoint.
+
+### Validate a configuration before import
+
+Provide the configuration JSON inline or via a file. The CLI wraps it into the `ValidateRequest` envelope for you.
+
+```
+content-cli asset-registry validate --assetType BOARD_V2 \
+  --packageKey my-pkg --configuration '{"components":[{"type":"kpi"}]}'
```
promeris: What was the reason for enabling configuration passing directly instead of using a file? I expect this to be very unusable even for an agent to do

docs/user-guide/asset-registry-commands.md
Buqeta: @ZgjimHaziri Can you do this fix? Otherwise, the other parts of the PR look good to me.

ZgjimHaziri: Removing the --configFile option fully. It seems overkill, considering we also have the --file option.

docs/user-guide/asset-registry-commands.md
ZgjimHaziri: My thought process here was more in this direction:
- The structure of the file that is used as a request is not defined anywhere (packageKey, nodeKey, configuration)
- So as an alternative for when we're not aware of the file structure: fully separated options for all fields in the transport. One for packageKey, one for nodeKey, and one for configuration.

promeris (dismissed):

Buqeta (dismissed):

Buqeta (approved):

PR #344, opened 2026-04-21
SP-38: Add new CLI commands for singe node operations

Buqeta:
#### Description

Adds new CLI commands for single-node operations against the Pacman public API:

- `config nodes create --packageKey ... [--body <json> | -f <file>] [--validate] [--json]` — creates a staging node. `--validate` calls the validate-only endpoint (no persistence) and exits with a success log if the payload is valid; otherwise fails.
- `config nodes update --packageKey ... --nodeKey ... [--body <json> | -f <file>] [--validate] [--json]` — updates a staging node. `--validate` behaves the same as above.
- `config nodes delete --packageKey ... --nodeKey ... [--force]` — archives a staging node.

Payload can be provided either inline via `--body` or from a file via `-f/--file` (mutually exclusive; one is required for create/update).

#### Relevant links

- Jira: https://celonis.atlassian.net/browse/SP-38
- Pacman endpoints PR: https://github.com/celonis/pacman/pull/1647

#### Checklist

### DIFF
--- a/src/commands/configuration-management/module.ts
+++ b/src/commands/configuration-management/module.ts
@@ -12,7 +12,6 @@ import { NodeDiffService } from "./node-diff.service";
 import { NodeDependencyService } from "./node-dependency.service";
 import { PackageVersionCommandService } from "./package-version-command.service";
 import { PackageValidationService } from "./package-validation.service";
-import { fileService } from "../../core/utils/file-service";
 
 class Module extends IModule {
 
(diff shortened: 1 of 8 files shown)


src/commands/configuration-management/module.ts
```diff
@@ -116,6 +117,32 @@ class Module extends IModule {
...
+            .description("Update a staging node in a package")
+            .requiredOption("--packageKey <packageKey>", "Identifier of the package")
+            .requiredOption("--nodeKey <nodeKey>", "Identifier of the node")
+            .option("--body <body>", "Node payload as JSON string")
+            .option("-f, --file <file>", "Path to a JSON file containing the node payload")
+            .option("--validate", "Only validate the payload without persisting. Returns success if valid.", false)
+            .option("--json", "Return the response as a JSON file")
+            .action(this.updateNode);
+
+        nodesCommand.command("delete")
```
promeris: Should we match the name of the API here? Also in the description would use only archive terminology
```suggestion
        nodesCommand.command("archive")
```

Buqeta: Changed it. Can you check?

promeris (dismissed):

src/commands/configuration-management/module.ts
```diff
@@ -249,6 +276,33 @@ class Module extends IModule {
...
+        const body = Module.resolveBody(options.body, options.file);
+        await new NodeService(context).createNode(options.packageKey, body, options.validate, options.json);
+    }
+
+    private async updateNode(context: Context, command: Command, options: OptionValues): Promise<void> {
+        const body = Module.resolveBody(options.body, options.file);
+        await new NodeService(context).updateNode(options.packageKey, options.nodeKey, body, options.validate, options.json);
+    }
+
+    private static resolveBody(body: string | undefined, file: string | undefined): string {
```
ZgjimHaziri: Ideally we would keep this out of the module.ts.

ZgjimHaziri (dismissed):

src/commands/configuration-management/module.ts
Buqeta: @ZgjimHaziri Moved it to the service file.

promeris (approved):

ZgjimHaziri (approved):

Buqeta:
Ignoring the coverage check, since it requires covering the transport files and the module.ts file, which doesn't make much sense.

PR #349, opened 2026-05-07
SP-160: add branch awareness for listing packages

alex-probst:
#### Description

This PR added branch awareness for the package listing `list` command by introducing the `--branches` option. 

Relevant requests are now forwarded to the Pacman API using the optional `includeBranches` parameter.

The `--branches` option comes up with a new validation requirement. Consumers will not be allowed to use `--branches` in combination with `--packageKeys` or `--keysByVersion` (justification: it makes logically not much sense to allow this combination at this moment).

#### Relevant links

- Jira: [SP-160](https://celonis.atlassian.net/browse/SP-160)
- Pacman PR: [PR 1799](https://github.com/celonis/pacman/pull/1799)

#### Checklist

[SP-160]: https://celonis.atlassian.net/browse/SP-160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/commands/configuration-management/module.ts
+++ b/src/commands/configuration-management/module.ts
@@ -187,16 +187,6 @@ class Module extends IModule {
             throw new Error("Please provide either --packageKeys or --keysByVersion, but not both.");
         }
 
-        if (options.branches) {
-            if (options.packageKeys) {
-                throw new Error("Please provide either --packageKeys or --branches, but not both.");
-            }
-
-            if (options.keysByVersion) {
-                throw new Error("Please provide either --keysByVersion or --branches, but not both.");
-            }
-        }
-
         await new ConfigCommandService(context).listPackages(
             options.json,
             options.flavors,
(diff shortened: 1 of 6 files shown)


promeris (dismissed):

src/commands/configuration-management/module.ts
```diff
@@ -185,7 +186,26 @@ class Module extends IModule {
         if (options.packageKeys && options.keysByVersion) {
             throw new Error("Please provide either --packageKeys or --keysByVersion, but not both.");
         }
-        await new ConfigCommandService(context).listPackages(options.json, options.flavors, options.withDependencies, options.packageKeys, options.keysByVersion, options.variableValue, options.variableType);
+
+        if (options.branches) {
```
LaberionAjvazi: We should not block these two commands - they are independent of the flag for find all. If branching is enabled clients should be able to list, export and import branches. If the branch specified by that key/version does not exist, we throw relevant errors. This is one of the requirements.

alex-probst: > We should not block these two commands - they are independent of the flag for find all. If branching is enabled clients should be able to list, export and import branches. If the branch specified by that key/version does not exist, we throw relevant errors. This is one of the requirements.

Fixed with a09015c8a7e97138c64e01248104fdc3f0b760f7.

LaberionAjvazi (approved):

PR #352, opened 2026-05-12
SP-143: add staging option to config list

kuvia:
#### Description

add `--staging` option to `config list`

#### Relevant links

https://celonis.atlassian.net/browse/SP-143

#### Checklist

### DIFF
--- a/src/commands/configuration-management/api/batch-import-export-api.ts
+++ b/src/commands/configuration-management/api/batch-import-export-api.ts
@@ -34,7 +34,7 @@ export class BatchImportExportApi {
         queryParams.set("includeBranches", includeBranches.toString());
         flavors.forEach(flavor => queryParams.append("flavors", flavor));
 
-        return this.httpClient().get(`/package-manager/api/core/staging/packages/export/list?${queryParams.toString()}`).catch(e => {
+        return this.httpClient().get(`/pacman/api/core/staging/packages/export/list?${queryParams.toString()}`).catch(e => {
             throw new FatalError(`Problem getting staging packages: ${e}`);
         });
     }
--- a/src/commands/configuration-management/api/batch-import-export-api.ts
+++ b/src/commands/configuration-management/api/batch-import-export-api.ts
@@ -21,10 +21,21 @@ export class BatchImportExportApi {
 
         queryParams.set("withDependencies", withDependencies.toString());
         queryParams.set("includeBranches", includeBranches.toString());
-        flavors.forEach(flavor => queryParams.append("flavors", flavor))
+        flavors.forEach(flavor => queryParams.append("flavors", flavor));
 
         return this.httpClient().get(`/package-manager/api/core/packages/export/list?${queryParams.toString()}`).catch(e => {
             throw new FatalError(`Problem getting active packages: ${e}`);
+        });
+    }
+
+    public async findAllStagingPackages(flavors: string[], includeBranches: boolean = false): Promise<PackageExportTransport[]> {
+        const queryParams = new URLSearchParams();
+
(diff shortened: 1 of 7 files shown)


tests/commands/configuration-management/config-list.spec.ts
```diff
@@ -275,4 +275,48 @@ describe("Config list", () => {
         expect(exportedSecondPackage).toEqual(secondPackage);
         expect(exportedFirstPackage).toEqual({...firstPackage, spaceId: "spaceId-1"});
     })
+
+    it("Should list all staging packages by key for non-json response with flavors", async () => {
```
promeris, line 279: let's also add tests when flavors not provided at all

src/commands/configuration-management/module.ts
```diff
@@ -183,6 +184,9 @@ class Module extends IModule {
     }
 
     private async listPackages(context: Context, command: Command, options: OptionValues): Promise<void> {
+        if (options.staging && (options.withDependencies || options.packageKeys || options.keysByVersion || options.variableValue || options.variableType)) {
+            throw new Error("Staging parameter is not compatible with --withDependencies, --packageKeys, --keysByVersion, --variableValue, --variableType");
```
Buqeta, line 188:
```suggestion
            throw new Error("Staging parameter is not compatible with --withDependencies, --packageKeys, --keysByVersion, --variableValue, --variableType. It's only compatible with --flavors (optional).");
```

Buqeta (commented):
Let's not forget to add the new endpoint to the OAuth allowlist [here](https://github.com/celonis/cfg-ibc/blob/master/gateway/components/opa-policies/staging/policies/service-policies/package_manager.rego), otherwise we won't be able to use this new command in staging. When the endpoint is ready to reach production, we should create a deployment ticket to whitelist the endpoint in the prod overlay too.

tests/commands/configuration-management/config-list.spec.ts
kuvia: added

src/commands/configuration-management/module.ts
kuvia: I would not add it because it's also compatible with `--json` and `--includeBranches`, feel like the description text would get a bit unwieldy otherwise..

src/commands/configuration-management/api/batch-import-export-api.ts
```diff
@@ -21,13 +21,24 @@ export class BatchImportExportApi {
...
         });
     }
 
+    public async findAllStagingPackages(flavors: string[], includeBranches: boolean = false): Promise<PackageExportTransport[]> {
+        const queryParams = new URLSearchParams();
+
+        queryParams.set("includeBranches", includeBranches.toString());
+        flavors.forEach(flavor => queryParams.append("flavors", flavor));
+
+        return this.httpClient().get(`/package-manager/api/core/staging/packages/export/list?${queryParams.toString()}`).catch(e => {
```
Buqeta: This should be 

```suggestion
        return this.httpClient().get(`/pacman/api/core/staging/packages/export/list?${queryParams.toString()}`).catch(e => {
```

kuvia: The tests for the prod endpoint also say `package-manager` should I change them too, then? (including the api client code in `batch-import-export-api.ts`)

Buqeta: No, we can leave them as is, since they're the old endpoints that were created before the pacman/package-manager split.

Buqeta (approved):

promeris (approved):

PR #356, opened 2026-05-22
SP-397: include description in asset-registry list output

ksalihu:
`asset-registry list` previously rendered each entry as `<assetType> - <displayName> [<group>]`, dropping the descriptor's `description` field. Append ` - <description>` when a description is present and leave the line unchanged for descriptors with `null` descriptions. Update the user-guide example output (also drops a stale `(basePath: ...)` suffix the docs claimed but the code never produced) and refresh the unit test to cover both the with-description and no-description branches.

Changed files:
- `src/commands/asset-registry/asset-registry.service.ts` — `logDescriptorSummary` now conditionally appends the description.
- `tests/commands/asset-registry/asset-registry-list.spec.ts` — covers both the with-description and no-description branches (asserts no trailing ` - ` and no `null` literal when the descriptor has no description).
- `docs/user-guide/asset-registry-commands.md` — example output refreshed and a one-liner added explaining the format. Also drops a stale `(basePath: ...)` suffix the docs claimed but the code never produced.

#### Testing

- `npx jest --testPathPattern="asset-registry/asset-registry-list"` — 3/3 passing
- Full `asset-registry` suite — 36/36 passing
- Built locally and ran `asset-registry list` against develop; all 8 registered asset types now show their description after the group.

#### Relevant links

- Jira: [SP-397](https://celonis.atlassian.net/browse/SP-397)

#### Checklist

Includes-AI-Code: true

[SP-397]: https://celonis.atlassian.net/browse/SP-397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/commands/asset-registry/asset-registry-list.spec.ts
+++ b/tests/commands/asset-registry/asset-registry-list.spec.ts
@@ -42,22 +42,44 @@ describe("Asset registry list", () => {
         },
     };
 
-    it("Should list all asset types with description when present", async () => {
-        mockAxiosGet("https://myTeam.celonis.cloud/pacman/api/core/asset-registry/types", metadata);
+    it("Should render the description when it is present", async () => {
+        const withDescription: AssetRegistryMetadata = {
+            types: {
+                SEMANTIC_MODEL: metadata.types.SEMANTIC_MODEL,
+            },
+        };
+        mockAxiosGet(
+            "https://myTeam.celonis.cloud/pacman/api/core/asset-registry/types",
+            withDescription
+        );
 
         await new AssetRegistryService(testContext).listTypes(false);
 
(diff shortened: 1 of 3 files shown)


tests/commands/asset-registry/asset-registry-list.spec.ts
```diff
@@ -42,17 +42,22 @@ describe("Asset registry list", () => {
         },
     };
 
-    it("Should list all asset types", async () => {
+    it("Should list all asset types with description when present", async () => {
```
promeris: We should test cases when description is present and not present.

Buqeta (approved):

promeris (approved):

PR #357, opened 2026-05-22
SP-347: add asset-registry skills list command

ksalihu:
### Summary

Adds content-cli asset-registry skills list to discover all agent skills registered in the asset registry (name, description, path), backed by GET /pacman/api/core/asset-registry/skills.

### Linked

https://celonis.atlassian.net/browse/SP-347

### What changed

- Introduced an asset-registry skills sub-command group so follow-up skill commands (get, file download, etc.) can slot in under the same tree.
- asset-registry skills list — prints one line per skill: <name> (<path>) - <description> (description omitted when empty); logs No agent skills registered. when the list is empty.
- asset-registry skills list --json — writes the full API response to a UUID-named JSON file in the working directory (same pattern as asset-registry list --json).
- Types: AgentSkill, AgentSkillMetadata, AgentSkillsResponse.
- User guide: new List Skills section in docs/user-guide/asset-registry-commands.md.

### Test plan

- [ ] content-cli asset-registry skills list --profile <develop> — verify non-empty output against develop realm
- [ ] content-cli asset-registry skills list --profile <develop> --json — verify file is written and contains skills[] with name, description, path, metadata.version


gramozkaragjyzi (dismissed):

promeris (dismissed):

src/commands/asset-registry/asset-registry-api.ts
```diff
@@ -18,6 +22,14 @@ export class AssetRegistryApi {
             });
     }
 
+    public async listSkills(): Promise<AgentSkillsResponse> {
+        return this.httpClient()
+            .get("/pacman/api/core/asset-registry/skills")
```
Buqeta, line 27: Let's not forget to add this to the AuthZ endpoints allowlist, if it's not there already.

Buqeta (dismissed):

src/commands/asset-registry/asset-registry-api.ts
ksalihu: yup

Buqeta (approved):

PR #362, opened 2026-05-28
SP-355: Implement diff with file

tneum-celonis:
#### Description

Extends `config nodes diff` to diff a local node JSON file against a node version stored in pacman, using the new public endpoint added in SP-144 (`POST /pacman/api/core/packages/{packageKey}/nodes/{nodeKey}/diff/configuration/with-file`).
- New `-f, --file <file>` flag, mutually exclusive with `--compareVersion`
- New `NodeDiffApi.diffWithFile(...)`
- `NodeDiffService` reuses the existing rendering (text + `--json`) for both flows via a shared private helper
- `DOCUMENTATION.md` updated

#### Relevant links

- Jira: https://celonis.atlassian.net/browse/SP-355

#### Checklist

### DIFF
--- a/src/commands/configuration-management/api/node-diff-api.ts
+++ b/src/commands/configuration-management/api/node-diff-api.ts
@@ -1,6 +1,11 @@
+import * as FormData from "form-data";
 import { HttpClient } from "../../../core/http/http-client";
 import { Context } from "../../../core/command/cli-context";
-import { GetNodeDiffRequest, NodeConfigurationDiffTransport } from "../interfaces/node-diff.interfaces";
+import {
+    GetNodeDiffRequest,
+    GetNodeDiffWithFileRequest,
+    NodeConfigurationDiffTransport,
+} from "../interfaces/node-diff.interfaces";
 import { FatalError } from "../../../core/utils/logger";
 
 export class NodeDiffApi {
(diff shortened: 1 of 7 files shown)


promeris (dismissed):

tests/commands/configuration-management/config-node-diff.spec.ts
```diff
@@ -290,4 +291,77 @@ describe("Node diff", () => {
         const changeDateLog = loggingTestTransport.logMessages.find(log => log.message.includes("Change Date:"));
         expect(changeDateLog.message).toContain(specificDate);
     });
+
+    describe("With file", () => {
```
ksalihu, line 322: Can you please add a few more tests, to reach >90% coverage, thats needed for the launch

ksalihu (changes requested):

tests/commands/configuration-management/config-node-diff.spec.ts
tneum-celonis: Done

src/commands/configuration-management/api/node-diff-api.ts
```diff
@@ -22,4 +27,21 @@ export class NodeDiffApi {
                 throw new FatalError(`Problem getting the node diff: ${exception}`);
             });
     }
+
+    public async diffWithFile(request: GetNodeDiffWithFileRequest): Promise<NodeConfigurationDiffTransport> {
+        const queryParams = new URLSearchParams();
+        queryParams.set("baseVersion", request.baseVersion);
+
+        const formData = new FormData();
+        formData.append("file", request.file);
```
ksalihu, line 36: Should there be a null check to pass only one of them based on the passed params of the cli? (not super sure how the API behaves if sent null/empty values here)

ksalihu (approved):

Buqeta (approved):

src/commands/configuration-management/api/node-diff-api.ts
tneum-celonis: A null value cannot get there because fs.createReadStream throws if a missing file arg somehow gets past the cli checks so I think this isn't needed

PR #363, opened 2026-05-28
SP-778: tighten content-cli SonarCloud coverage scope

ksalihu:
#### Description

Today the SonarCloud coverage metric for `content-cli` is artificially low (~52% statements / ~72% functions) because it counts files where unit-testing isn't possible or meaningful. This PR scopes the coverage measurement to source files that can actually be tested, without hiding files that have real logic but currently lack tests.

**What's excluded from coverage** (added to `sonar.coverage.exclusions` + mirrored in `jest.config.ts` `collectCoverageFrom`):

- `**/*.interface.ts`, `**/*.interfaces.ts` — pure type / interface declarations
- `**/*.constants.ts` — enum and constant maps
- `src/content-cli.ts` — CLI entry point (top-level `program.parse()`, `process.exit`, `uncaughtException` handler — not unit-testable without spawning a subprocess)

These files remain in SonarCloud analysis for bugs / smells / hotspots — only the coverage calculation skips them.

**What deliberately stays in coverage** (even though some have 0% today):

- `**/*-factory.ts` / `**/*.factory.ts` — 7 of the 8 factories contain real FS operations, branching, and validation paths (`widget.manager-factory.ts` alone has 5 manifest-validation branches that throw `FatalError`). None have tests today. Surfacing them as gaps is the point.
- `**/module.ts` — most have untested `register()` wiring. `configuration-management/module.ts` also has real handler-level validation logic that should keep being tracked.

**Sonar config cleanup**:

- `**/*.test.ts` references removed from `sonar.exclusions` and `sonar.test.inclusions` — the project uses `.spec.ts` only (`jest.testMatch: <rootDir>/tests/**/*.spec.ts`).
- `**/*.d.ts` and `dist-bundle/**` added to `sonar.exclusions`.
- Trailing-comma / dangling line-continuation in the exclusions list cleaned up.

**New `register()` smoke-test pattern**:

To avoid permanently excluding `module.ts` files (whose Commander wiring is otherwise hard to cover), this PR adds a small reusable chainable Configurator mock at `tests/utls/configurator-mock.ts` and applies it to `tests/commands/configuration-management/module.spec.ts`. Calling `module.register(testContext, mockConfigurator)` exercises every `.command().option().action()` line in the file. Result for that module: **40.30% → 92.61% line coverage**. Other modules can adopt the same pattern as they're touched.

**Aggregate coverage**:

| Metric | Before | After |
|---|---|---|
| Statements | 52.13% | 55.58% |
| Branches | 86.83% | 87.97% |
| Functions | 71.65% | 74.00% |
| Lines | 52.13% | 55.58% |

The numbers are higher than baseline but lower than they would be if factories and module wiring were hidden. This is intentional: the metric now reflects real testable code rather than being inflated by hiding files that should have tests.

#### Relevant links

- Jira: https://celonis.atlassian.net/browse/SP-778

#### Checklist

#### Test plan

- `npx jest --coverage` passes (355/355 tests, was 353/353; +2 from the new `register()` smoke test)
- LCOV report at `coverage/lcov.info` contains source files only — verified that no `*.interface.ts` / `*.interfaces.ts` / `*.constants.ts` / `content-cli.ts` lines appear in the LCOV output
- Manually compared per-file coverage rows before/after to confirm the configuration-management module jumped to 92.61% and the previously-hidden factories now surface at 0%

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -15,12 +15,21 @@ sonar.sourceEncoding=UTF-8
 # Coverage reporting (from Jest / ts-jest)
 sonar.javascript.lcov.reportPaths=coverage/lcov.info
 
-# Exclusions (optional, adjust as needed)
-# Ignore built files and configs
-sonar.exclusions=dist/**, node_modules/**, coverage/**, **/*.test.ts
+# Build output and generated artifacts — excluded from all analysis
+sonar.exclusions=dist/**, dist-bundle/**, node_modules/**, coverage/**, **/*.d.ts
 
-# Test file inclusions
-sonar.test.inclusions=**/*.test.ts, **/*.spec.ts
+# Excluded from coverage only — still scanned for bugs and code smells.
+# These files contain no executable logic that unit tests can meaningfully cover:
+#   - *.interface.ts / *.interfaces.ts : type-only declarations
+#   - *.constants.ts                   : enum / constant maps
+#   - src/content-cli.ts               : CLI entry point (top-level execution)
+sonar.coverage.exclusions=src/content-cli.ts, \
+    **/*.interface.ts, \
+    **/*.interfaces.ts, \
+    **/*.constants.ts
+
+# Test file inclusions (only .spec.ts is used in this repo)
+sonar.test.inclusions=**/*.spec.ts
 
 # Report verbose output
 sonar.verbose=true
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -18,11 +18,17 @@ const config: Config.InitialOptions = {
         "<rootDir>/tests/jest.setup.ts",
     ],
     // Coverage configuration
+    // Keep the negation list in sync with sonar.coverage.exclusions in sonar-project.properties
+    // so local `jest --coverage` matches what SonarCloud counts.
     collectCoverageFrom: [
         "src/**/*.{ts,tsx}",
         "!src/**/*.d.ts",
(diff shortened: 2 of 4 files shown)


sonar-project.properties
```diff
@@ -15,12 +15,21 @@ sonar.sourceEncoding=UTF-8
...
+sonar.exclusions=dist/**, dist-bundle/**, node_modules/**, coverage/**, **/*.d.ts
 
-# Test file inclusions
-sonar.test.inclusions=**/*.test.ts, **/*.spec.ts
+# Excluded from coverage only — still scanned for bugs and code smells.
+# These files contain no executable logic that unit tests can meaningfully cover:
+#   - *.interface.ts / *.interfaces.ts : type-only declarations
+#   - *.constants.ts                   : enum / constant maps
+#   - src/content-cli.ts               : CLI entry point (top-level execution)
+sonar.coverage.exclusions=src/content-cli.ts, \
```
Buqeta, line 26: Can we exclude `md` files from coverage calculations?

ksalihu: the sonar.source is in src directory, which means that the docs are not included, so no md files are calculated

Buqeta (approved):

ksalihu:
Merging to see if this fixes the coverage report. Feel free to add comments, will address them as needed in follow up PR

PR #369, opened 2026-06-04
SP-874: Add single package import command

ZgjimHaziri:
#### Description

Adds a new `config package import` command for importing a single package — the CLI counterpart to the new pacman single-package import API (`POST /pacman/api/core/staging/packages/import-file`).

How it was done:

- Added it under a new `config package` subgroup instead of overloading the existing batch `config import`. The two use different, **non-interchangeable** archive formats (the batch archive has nested per-package zips + manifest; the single-package format is a flat `package.json` + `nodes/`), so keeping them separate avoids confusing the workflows.
- The command imports from a package zip (`--file`) or a directory (`--directory`, which the CLI flat-zips before upload), with an `--overwrite` flag and optional `--json` output. A 4 GB uncompressed guard is enforced on the archive.
- Reframed the config docs and CLI help so the package commands read as the default for working with a package, and the batch commands are presented as their own batch-specific set whose artifacts don't mix with the single-package format.

#### Relevant links

- Jira: https://celonis.atlassian.net/browse/SP-874
- API PR: celonis/pacman#1818

#### Checklist

- [x] I have tested the change and proved that it works in different scenarios (unit tests cover file/directory imports, overwrite, JSON output, and the validation/error paths; full suite green)

Includes-AI-Code: true

### DIFF
--- a/docs/user-guide/config-commands.md
+++ b/docs/user-guide/config-commands.md
@@ -15,7 +15,7 @@ The batch commands are **batch-specific**: they use their own archive format tha
 
 ## Two command families
 
-The batch commands are a **self-contained, batch-specific set**. `config export` produces a multi-package **batch archive** — a top-level `manifest.json`, `variables.json`, `studio.json`, and one nested `<packageKey>_<version>.zip` per package — that is produced and consumed **only** by other batch commands. `config package import`, by contrast, works with a plain **package zip** (a `package.json`, an optional `variables.json`, and a `nodes/` folder). The two formats are **not interchangeable**:
+The batch commands are a **self-contained, batch-specific set**. `config export` produces a multi-package **batch artifact** — a top-level `manifest.json`, `variables.json`, `studio.json`, and one nested `<packageKey>_<version>.zip` per package — that is produced and consumed **only** by other batch commands. `config package import`, by contrast, works with a plain **package zip** (a `package.json`, an optional `variables.json`, and a `nodes/` folder). The two formats are **not interchangeable**:
 
 - An archive from `config export` can be imported with `config import` or inspected with `config diff` — but **not** with `config package import`.
 - A package zip used by `config package import` **cannot** be imported with `config import` or diffed with `config diff`.
@@ -105,7 +105,7 @@ content-cli config list -p <sourceProfile> --packageKeys key1 ... keyN
 
 ## Batch Export Packages
 
-> **Batch command.** Part of the [batch family](#two-command-families). It produces a multi-package **batch archive** that can only be re-imported with [`config import`](#batch-import-packages) or inspected with [`config diff`](#diff-local-zip-with-deployed-versionspecific-versionstaging) — **not** with `config package import`. To work with one package, use [`config package import`](#package-commands-config-package).
+> **Batch command.** Part of the [batch family](#two-command-families). It produces a multi-package **batch artifact** that can only be re-imported with [`config import`](#batch-import-packages) or inspected with [`config diff`](#diff-local-zip-with-deployed-versionspecific-versionstaging) — **not** with `config package import`. To work with one package, use [`config package import`](#package-commands-config-package).
 
 Packages can be exported using the following command:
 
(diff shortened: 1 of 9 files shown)


src/commands/configuration-management/module.ts
```diff
@@ -60,6 +62,17 @@ class Module extends IModule {
             .option("-d, --directory <directory>", "Exported packages directory (relative path)")
             .action(this.batchImportPackages);
 
+        const packageCommand = configCommand.command("package")
+            .description("Commands for working with a package.");
+
+        packageCommand.command("import")
+            .description("Import a package from a zip file or directory. Uses the package format, which is not interchangeable with the batch 'config export' / 'config import' archive.")
+            .option("-f, --file <file>", "Package zip file (relative path)")
+            .option("-d, --directory <directory>", "Package directory (relative path)")
```
ksalihu, line 71: Something that could make the interface a bit simpler, could be a `source` paramter that replaces both file and directory, and could be either a zip file or a directory and is resolved internally by the tool to determine its type

ZgjimHaziri: I like the explicit options a bit more, since it's more clear what the options are. Let's see what the others think, and we can decide based on that.

docs/user-guide/config-commands.md
```diff
@@ -177,6 +213,81 @@ content-cli config import -p <sourceProfile> -d <export_dir> --validate --overwr
...
 
+## Package Commands (`config package`)
+
+The `config package` command group works with a package and its contents. It is **not** part of the batch-specific set — it uses the plain package format described below, which is not interchangeable with the batch archive (see [Two command families](#two-command-families)).
+
+### Import a Package
+
+`config package import` imports a package from a package zip (or directory). Unlike [`config import`](#batch-import-packages) — which performs a **batch** import and expects the multi-package batch archive (`manifest.json`, a top-level `variables.json`, `studio.json`, and a nested `<packageKey>_<version>.zip` per package) — `config package import` takes a plain, flat package layout and imports it on its own.
+
+> A zip produced by `config export` is a **batch archive** and cannot be imported with `config package import`. Likewise, a package zip cannot be imported with `config import`. Use the command that matches how the artifact was produced.
```
Buqeta: > A zip produced by `config export` is a **batch archive** and cannot be imported with `config package import`.

I'm not sure I understand what this means? Can you explain?

docs/user-guide/config-commands.md
```diff
@@ -177,6 +213,81 @@ content-cli config import -p <sourceProfile> -d <export_dir> --validate --overwr
...
+
+#### Overwriting an Existing Package
+
+By default the import fails if a package with the same key already exists. Use `--overwrite` to replace it:
+
+```bash
+content-cli config package import -p <sourceProfile> -f <file path> --overwrite
+```
+
+When overwriting, variable assignments whose key is no longer declared in the imported `package.json` are pruned, keeping declarations and assignments consistent.
```
Buqeta: > When overwriting, variable assignments whose key is no longer declared in the imported `package.json` are pruned

Can we say `ingored`, instead of `pruned`?

src/commands/configuration-management/module.ts
```diff
@@ -60,6 +62,17 @@ class Module extends IModule {
             .option("-d, --directory <directory>", "Exported packages directory (relative path)")
             .action(this.batchImportPackages);
 
+        const packageCommand = configCommand.command("package")
+            .description("Commands for working with a package.");
```
Buqeta:
```suggestion
            .description("Commands for working with a single package.");
```

promeris (dismissed):

docs/user-guide/config-commands.md
ZgjimHaziri: Batch **artifact**, would be the right word. This means that if you export via config export, you cannot import via config package import. Because the structure of the former batch one, is different from the latter single import.

Buqeta: Let's change it then.

Buqeta (approved):

src/commands/configuration-management/module.ts
ksalihu: Okay, sounds good to me

ksalihu (approved):

PR #370, opened 2026-06-05
SP-957: move Team-to-Team Copy commands to 't2tc package' group

ZgjimHaziri:
#### Description

The batch commands `config list/export/import/diff` are specific to the Team-to-Team Copy (T2TC) service but lived under the generic top-level `config` group, mixed in with the package-resource commands (`package`, `nodes`, `versions`, `variables`, `metadata`). This made the surface confusing: bulk T2TC-archive operations looked like general-purpose commands.

**Approach:** group the T2TC operations under a dedicated `t2tc package` group, and keep `config` for package-resource commands only.

- `config list/export/import/diff` → `t2tc package list/export/import/diff`
- `config validate` → `config package validate` (it validates a package)
- Added `config package list` for everyday package listing (staging packages by default, with flavor filtering)

**Non-breaking:** the old commands are kept as deprecated aliases. Each prints a runtime deprecation notice pointing to its replacement and is marked `[Deprecated]` in `--help`, so existing automation keeps working until the aliases are removed in a future release.

**Internal structure:** the T2TC logic now lives in a dedicated `t2tc` module (command/package services, package + diff APIs, `StudioService`, transport constants), so the deprecated `config` aliases are thin delegators. The former `BatchImportExport` catch-all in `config` is split into domain-specific classes — `StagingPackage` (staging list), `Metadata` (package metadata), `Variable` (variable export) — meaning `configuration-management` no longer carries any batch export/import code.

Docs updated accordingly (new `t2tc-commands.md`, reframed `config-commands.md`, mkdocs nav).

#### Relevant links

- Jira: https://celonis.atlassian.net/browse/SP-957

#### Checklist


src/commands/configuration-management/module.ts
```diff
@@ -193,6 +216,53 @@ class Module extends IModule {
             .option("--json", "Return the response as a JSON file")
             .action(this.listNodeDependencies);
 
+        const t2tcCommand = configurator.command("t2tc")
```
promeris: can we create a different module for this?

Buqeta (approved):

docs/user-guide/t2tc-commands.md
```diff
@@ -0,0 +1,211 @@
...
+exported_package_random_uuid/
+├─ manifest.json
+├─ variable.json
+├─ studio.json
+├─ package_key1-version.zip
+├─ ...
+├─ package_keyN-version.zip
+```
+
+### Git Integration for Export
```
promeris, line 109: This needs to be supported for the single export/import commands now as well. Let's make sure to plan this accordingly

ZgjimHaziri: Created a ticket for it: https://celonis.atlassian.net/browse/SP-992.

src/commands/configuration-management/module.ts
ZgjimHaziri: Created it and refactored the services/apis in which module they fit most.

Buqeta (dismissed):

Buqeta (approved):

promeris (approved):

PR #373, opened 2026-06-10
SP-1089: Add plan for command integration tests

tneum-celonis:
DO NOT MERGE

Plan for [SP-1089](https://celonis.atlassian.net/browse/SP-1089).

[SP-1089]: https://celonis.atlassian.net/browse/SP-1089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #374, opened 2026-06-10
SP-1089: Command integration tests

tneum-celonis:
#### Description

Based on the plan in #373.

Drives Content CLI command tests through the real Commander parser instead of poking private action methods. The two existing "module" specs bypassed Commander's command registration, option parsing, defaults, type coercion, and validation. After this change, tests mirror what an end user types and run through the same parse path the `content-cli` binary uses.

- `src/content-cli.ts`:
  - exposes a command factory and hides `run()` behind a module check to prevent running in test mode
  - add explicit providing of modules as an alternative to automatic discovery to facilitate testing / mocking
- `tests/utls/cli-program.ts`: Wires `testContext` into `createProgram` and skips filesystem-based module discovery (which would conflict with fs mocks)
- Move and rewrite the two module tests
  - Move test files to an integration folder for clarity
  - Invoke CLI through `program.parseAsync`
- Validate logger output instead of working with the result of `parse` (now `parseAsync`)

#### Relevant links

- Jira: [SP-1089](https://celonis.atlassian.net/browse/SP-1089)

#### Checklist

[SP-1089]: https://celonis.atlassian.net/browse/SP-1089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kuvia (dismissed):

celeszter (dismissed):

src/content-cli.ts
```diff
@@ -14,74 +14,109 @@ import { ContentCLIHelp } from "./core/command/CustomHelp";
...
 
-    const moduleHandler = new ModuleHandler(program, context);
+    // Parse global options up-front so banner/debug-level decisions can use
+    // them before module discovery runs.
+    const bootstrapProgram = new Command();
+    bootstrapProgram.option("-q, --quietmode", "Reduce output to a minimum", false);
+    bootstrapProgram.option("-p, --profile [profile]");
+    bootstrapProgram.option("--gitProfile [gitProfile]", "Git profile to use");
+    bootstrapProgram.option("--debug", "Print debug messages", false);
+    bootstrapProgram.option("--dev", "Development Mode", false);
```
ksalihu: Could you pull these five global-option declarations into a shared helper that both run's bootstrap and createProgram call? As is, every new root flag has to be added in two places to stay in sync.

Separately — were --quietmode, --debug, --dev, and --profile <x> <subcommand> checked manually through run()? The integration tests build the program via createProgram directly, so I couldn't find where this bootstrap path gets exercised and want to make sure it's covered before merge.

kuvia: What do you mean by the second paragraph? Whether the flags have been tested by running content-cli in the terminal manually? Quiet mode and debug are evaluated just slightly below

ksalihu: I meant are they covered from tests or tested manually

kuvia: As we didn't have these high level tests before we don't have a regression anyway so I tested this manually. Also removed the duplication

ksalihu (approved):

PR #375, opened 2026-06-11
SP-1090: Assert on real process output and exit codes in integration tests

admirjashari:
#### Description

**Changes**
- `module-handler.ts`: the shared command action wrapper now sets `process.exitCode = 1` when a command throws, so errors surface as a non-zero exit code (previously thrown errors were only logged and the process still exited `0`). `GracefulError` is preserved as exit `0`.
- `package.manager.ts`: fixed a bare `process.exit()` (exited `0`) on an error branch → `process.exit(1)`.

**Tests Added**
- Added `tests/utls/cli-runner.ts` (`runCli`), which drives a command through the real Commander wiring — reusing the existing `buildTestProgram` helper — and captures the actual `stdout`, `stderr`, and exit code.
- Reworked the existing integration specs to run through `runCli` and assert on **real captured output + exit codes** instead of the injected logger transport:
  - `configuration-management.spec.ts`: validation errors now assert real output + non-zero exit code (replacing the in-memory winston transport checks).
  - `asset-registry.spec.ts` (the agentic-dev surface): each command asserts it exits `0` on success; error paths assert real output + exit code — service failure → exit `1`, unknown command / missing required option → non-zero, and the `GracefulError` (exit `0`) vs. regular-error (exit `1`) contract.

#### Relevant links

- Jira: https://celonis.atlassian.net/browse/SP-1090

#### Checklist


tests/integration/cli-process-output.spec.ts
```diff
@@ -180,4 +185,49 @@ describe("CLI process output and exit codes", () => {
             expect(result.exitCode).toBe(1);
         });
     });
+
+    describe("config export", () => {
+        it("downloads zip and exits 0", async () => {
```
kuvia: I'm confused, you added new tests but I understood that the existing integration tests in `integration/commands/asset-registry.spec.ts` and `integration/commands/configuration-management.spec.ts` should be enhanced by asserting on stdout. They also cover a lot more cases than those three here.

tests/integration/commands/asset-registry.spec.ts
kuvia: I think instead of adding new tests the existing tests should use your `runCliProcess`. You probably don't need to assert on the exit code in every test but the output is different for every command probably.

tests/integration/cli-process-output.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+import { Configurator, IModule } from "../../src/core/command/module-handler";
+import { Context } from "../../src/core/command/cli-context";
+import { FatalError, GracefulError } from "../../src/core/utils/logger";
+import { VersionUtils } from "../../src/core/utils/version";
+
+import AssetRegistryModule = require("../../src/commands/asset-registry/module");
+
+const GRACEFUL_MESSAGE = "graceful failure - should not fail the process";
+
+class DiagnosticsModule extends IModule {
```
kuvia: This class isn't needed as it's just example tests once you've updated the other ones we are covering most of the agentic commands.

Buqeta (approved):

PR #376, opened 2026-06-15
SP-1091: replace json writes and zip file reads with real files in temp directory

kuvia:
#### Description

Removed general mocking of `fs`, this had a lot of cascading changes, most of which are writes to JSON files that have all been replaced by real file writes. I have also replaced all mocks of zip file reading occasions in package import.

Instances where file mocks were kept are git branch and profile reads because they strictly don't read from a temp dir but from a git repo and homedir respectively. If we want to change those too I would propose to do it in a different task because the PR is big enough as is.

#### Relevant links

- Jira: https://celonis.atlassian.net/browse/SP-1091

#### Checklist


src/core/utils/file-service.ts
```diff
@@ -16,7 +16,7 @@ export class FileService {
         });
     }
 
-    public async readFileToJson<T>(fileName: string): Promise<T> {
+    public readFileToJson<T>(fileName: string): Promise<T> {
```
ksalihu: Can you also change the type and also not await it in the usages

tests/commands/configuration-management/config-list-variables.spec.ts
```diff
@@ -154,19 +148,15 @@ describe("Config listVariables", () => {
...
 
         expect(loggingTestTransport.logMessages.length).toBe(1);
         expect(loggingTestTransport.logMessages[0].message).toContain(FileService.fileDownloadedMessage);
 
-        const expectedFileName = loggingTestTransport.logMessages[0].message.split(FileService.fileDownloadedMessage)[1];
-        expect(mockWriteFileSync).toHaveBeenCalledWith(path.resolve(process.cwd(), expectedFileName), JSON.stringify(fixedVariableManifests), {encoding: "utf-8", mode: 0o600});
-
-        const variableExportRequest = parse(mockedPostRequestBodyByUrl.get("https://myTeam.celonis.cloud/package-manager/api/core/packages/export/batch/variables-with-assignments"));
-        expect(variableExportRequest).toEqual(packageKeyAndVersionPairs);
+        const downloadedFile = getJsonFromDownloadedFile();
```
merttuncay: Could we remove this variable assignment since it is not used and the function is directly called in the expect statement? Thanks!

tests/utls/fs-utils.ts
```diff
@@ -0,0 +1,38 @@
+import { readFileSync, mkdirSync } from "fs";
```
merttuncay: I think these imports should be from "node:fs", right? Thanks!

merttuncay (changes requested):

tests/commands/configuration-management/config-list-variables.spec.ts
kuvia: done, also replaced all occurrences of variable assignments where the full response body is asserted

tests/utls/fs-utils.ts
kuvia: It seems like we use a mix all over the place which is not great. I don't want to change them everywhere because that'll mean I'll need approvals from all teams involved but I've changed it to `node:fs` in all the places where I had changes.

ksalihu (dismissed):

merttuncay (dismissed):

ksalihu (dismissed):

ksalihu (approved):

merttuncay (approved):

PR #385, opened 2026-06-24
SP-348: Add command to retrieve single skill files

tneum-celonis:
#### Description

Adds a new `asset-registry skills get` command for downloading skill files registered in the asset registry. By default it fetches `SKILL.md` for the given `--path`; pass `--file <relative>` to retrieve any reference file bundled with the skill, and `--output <dir>` to choose a destination directory (defaults to cwd).

Also refactors a few small bits along the way:
- Extracts a shared `trimSlashes` helper in `src/core/utils/path.ts` (used by both the service and the URL-building code).
- `FileService.writeBufferToPath` now reuses `writeBufferToFileWithGivenName` for the actual write, and a new private `ensureDir` helper replaces every `fs.mkdirSync(..., { recursive: true, mode: ... })` call in the file.

#### Relevant links

- Jira: [SP-348](https://celonis.atlassian.net/browse/SP-348)

#### Checklist

[SP-348]: https://celonis.atlassian.net/browse/SP-348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/user-guide/asset-registry-commands.md
+++ b/docs/user-guide/asset-registry-commands.md
@@ -149,6 +149,71 @@ Options:
 - `--assetType <assetType>` (required) – The asset type identifier
 - `--json` – Write the examples to a JSON file in the working directory
 
+## Skills
+
+The asset registry also publishes agent skills (authored guidance for the platform and for specific asset types). Each skill exposes a `SKILL.md` and optional reference files.
+
+### List Skills
+
+List all skills available on the platform.
+
+```
+content-cli asset-registry skills list
+```
+
+Example output:
+
+```
+content-cli-setup (platform/content-cli-setup) - Install content-cli and create a profile against a Celonis team.
+asset-studio-board-v2 (asset/BOARD_V2/asset-studio-board-v2) - Authoring one Celonis Studio view asset of type BOARD_V2.
+```
+
+Each line is `<name> (<path>)` followed by ` - <description>` when the skill provides one. The `<path>` value is what you pass to `skills get --path`.
+
+Use `--json` to write the full response to a JSON file in the working directory:
+
+```
+content-cli asset-registry skills list --json
+```
+
+### Download a Skill File
+
+Download a skill's `SKILL.md` (or a specific reference file) to the local filesystem. The Studio MCP server remains the recommended source for live agent use; this command is a fetch/inspect utility for environments without the MCP server, for offline review, or for vendoring a copy into a repo.
+
+Download the default `SKILL.md` for a platform skill:
+
+```
+content-cli asset-registry skills get --path platform/content-cli-setup
+```
+
+Download a `SKILL.md` for an asset skill:
+
+```
+content-cli asset-registry skills get --path asset/BOARD_V2/asset-studio-board-v2
+```
+
+Download a specific reference file and write into a target directory:
+
+```
+content-cli asset-registry skills get \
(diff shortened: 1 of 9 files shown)


tests/commands/asset-registry/asset-registry-skills-get.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+import { loggingTestTransport } from "../../jest.setup";
+import { FatalError } from "../../../src/core/utils/logger";
+import { FileService } from "../../../src/core/utils/file-service";
+
+const SKILLS_BASE_URL = "https://myTeam.celonis.cloud/pacman/api/core/asset-registry/skills";
+
+describe("Asset registry skills get", () => {
+    const skillContent = Buffer.from("# Hello SKILL\n\nLine 2.\n", "utf-8");
+
+    function uniqueOutputDir(): string {
```
kuvia: the same helper functions exist in `fs-utils.ts`

src/commands/asset-registry/asset-registry-api.ts
```diff
@@ -49,4 +50,27 @@ export class AssetRegistryApi {
...
+            .catch((e) => handleAssetRegistryApiError(operation, e));
+    }
+
+    private buildSkillFileUrl(skillPath: string, filePath?: string): string {
+        const skillSegments = encodePathSegments(skillPath);
+        const base = `/pacman/api/core/asset-registry/skills/${skillSegments}`;
+        if (!filePath) {
+            return base;
+        }
+        return `${base}/${encodePathSegments(filePath)}`;
```
kuvia: I think except for the last trailing slash of the full url, `path.join` can actually do the job itself?

tests/commands/asset-registry/asset-registry-skills-get.spec.ts
tneum-celonis: Not really, there is makeTempDir but that's different. I will move this method to fs-utils though

src/commands/asset-registry/asset-registry-api.ts
tneum-celonis: `path.join` does more than we want here imo because it also collapses `..`.

ZgjimHaziri (approved):

docs/user-guide/asset-registry-commands.md
```diff
@@ -149,6 +149,71 @@ Options:
...
+```
+
+### Download a Skill File
+
+Download a skill's `SKILL.md` (or a specific reference file) to the local filesystem. The Studio MCP server remains the recommended source for live agent use; this command is a fetch/inspect utility for environments without the MCP server, for offline review, or for vendoring a copy into a repo.
+
+Download the default `SKILL.md` for a platform skill:
+
+```
+content-cli asset-registry skills get --path platform/content-cli-setup
```
ksalihu, line 186: What exactly does this path refer to? is it the path on how the files are stored? If so, we shouldn't couple the API to this at all

ksalihu (changes requested):

docs/user-guide/asset-registry-commands.md
tneum-celonis: It's the path of the skill on the backend. We're not coupling anything that isn't already coupled

ksalihu (approved):

PR #393, opened 2026-07-09
SP-1383: gate early-access commands behind feature flag, replace beta

admirjashari:
#### Description

Gate early-access (non-GA) commands behind a backend feature flag. If a team runs a gated command that isn't enabled for them, they get a clear "not enabled — contact support" message instead of a generic 403.

Also consolidates the old `beta` marker into a single `earlyAccess`:
- `earlyAccess(featureKey?)` replaces `beta()`/`betaOption()` — with a key it's gated at run time, without one it's just an internal marker.
- No `(beta)` label anymore; early-access status is internal (nothing shown in `--help`).
- Check uses the existing `/api/team/features` endpoint with the profile token (flag enabled = key present).
- No profile → normal "no profile" error; check failure → fail-closed "could not verify".
- `deployment` commands migrated to `.earlyAccess()` (ungated, behavior unchanged).

#### Relevant links

- Jira: https://celonis.atlassian.net/browse/SP-1383

#### Checklist

- [ ] Docs — N/A, tracked in a separate ticket

### DIFF
--- a/src/commands/deployment/module.ts
+++ b/src/commands/deployment/module.ts
@@ -6,11 +6,10 @@ import { DeploymentService } from "./deployment.service";
 class Module extends IModule {
 
     public register(context: Context, configurator: Configurator): void {
-        const deploymentCommand = configurator.command("deployment").beta()
+        const deploymentCommand = configurator.command("deployment")
             .description("Create deployments, list their history, check active deployments, and retrieve deployables and targets");
 
         deploymentCommand.command("create")
-            .beta()
             .description("Create a new deployment")
             .requiredOption("--packageKey <packageKey>", "Identifier of the package to deploy")
             .requiredOption("--packageVersion <packageVersion>", "Version of the package to deploy")
@@ -20,10 +19,9 @@ class Module extends IModule {
             .action(this.createDeployment);
 
         const listCommand = deploymentCommand.command("list")
-            .description("List deployment history, active deployments, deployables or targets").beta();
+            .description("List deployment history, active deployments, deployables or targets");
 
         listCommand.command("history")
-            .beta()
             .description("List deployment history")
             .option("--packageKey <packageKey>", "Filter deployment history by package key")
             .option("--targetId <targetId>", "Filter deployment history by target ID")
@@ -36,7 +34,6 @@ class Module extends IModule {
             .action(this.listDeploymentHistory);
 
         listCommand.command("active")
-            .beta()
             .description("Get the active deployment(s) for a given target or package.\n"+
                 "You can use the command to list the active deployment(s) for a specific target or for a specific package.\n" +
                 "The targetIds filter is available only for getting the active deployments for a given package. \n" +
@@ -50,14 +47,12 @@ class Module extends IModule {
             .action(this.listActiveDeployments);
 
         listCommand.command("deployables")
-            .beta()
             .description("List all deployables")
             .option("--flavor <flavor>", "Filter deployables by flavor")
             .option("--json", "Return the response as a JSON file")
             .action(this.listDeployables);
 
         listCommand.command("targets")
-            .beta()
             .description("List all targets for a given deployable type and package key")
             .requiredOption("--deployableType <deployableType>", "The type of the deployable")
             .requiredOption("--packageKey <packageKey>", "Identifier of the package to list targets for")
--- a/src/core/command/module-handler.ts
+++ b/src/core/command/module-handler.ts
@@ -3,6 +3,7 @@ import * as fs from "fs";
 import { Command, CommandOptions, Option, OptionValues } from "commander";
 import { Context } from "./cli-context";
 import { GracefulError, logger } from "../utils/logger";
+import { isFeatureDisabledError } from "../feature-flag/feature-disabled-error";
 import * as chalk from "chalk";
 
 export abstract class IModule {
@@ -129,8 +130,8 @@ export class Configurator {
     public rootCommandMap = new Map<string, CommandConfig>();
 
     constructor(
-        private program: Command,
-        private ctx: Context
+        private readonly program: Command,
+        private readonly ctx: Context
     ) {}
 
     /**
@@ -157,8 +158,8 @@ export class CommandConfig {
     private deprecationMessage: string;
 
     constructor(
-        private cmd: Command,
-        private ctx: Context
+        private readonly cmd: Command,
+        private readonly ctx: Context
     ) {}
 
     public command(nameAndArgs: string, opts?: CommandOptions): CommandConfig {
@@ -218,6 +219,14 @@ export class CommandConfig {
             } catch (error) {
                 if (error instanceof GracefulError) {
                     logger.error(error.message);
+                    return;
+                }
+                // Backend gates early-access features; translate its "feature disabled"
+                // rejection into a clear message instead of a raw error.
+                if (isFeatureDisabledError(error)) {
+                    logger.error(
+                        `'${this.cmd.name()}' is not enabled for your team. Contact support to request access.`
+                    );
                     return;
                 }
                 logger.error(`An unexpected error occured executing a command: ${error}`);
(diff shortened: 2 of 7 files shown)


src/core/command/module-handler.ts
```diff
@@ -203,17 +196,25 @@ export class CommandConfig {
...
+     * Marks a command as early access (non-GA). Purely internal — it is not shown
+     * in help. When a feature key is provided, the command is gated at run time:
+     * before it executes, the CLI checks whether that backend feature flag is
+     * enabled for the team, and stops with a clear message if it is not.
+     */
+    public earlyAccess(featureKey?: string): CommandConfig {
+        (this.cmd as any).isEarlyAccess = true;
+        if (featureKey) {
+            (this.cmd as any).earlyAccessFeatureKey = featureKey;
+        }
```
tneum-celonis: Nitpicky suggestion for readability, totally optional

```suggestion
        const command = this.cmd as any
        command.isEarlyAccess = true;
        if (featureKey) {
            command.earlyAccessFeatureKey = featureKey;
        }
```

src/core/command/module-handler.ts
```diff
@@ -232,22 +233,36 @@ export class CommandConfig {
...
-                logger.info(chalk.yellow(`The option '${option.long}' is in beta and may change in future releases.`));
+
+        let enabled: boolean;
+        try {
+            enabled = await new FeatureFlagService(this.ctx).isEnabled(featureKey);
+        } catch (error) {
+            // With no profile, surface the standard "no profile provided" error
+            // rather than masking it with an early-access message.
+            if (!this.ctx.profile) {
+                throw error;
```
tneum-celonis: This just rethrows the error emitted by the FF service, how does this "surface the standard 'no profile provided' error"?

admirjashari: With no profile, context.httpClient throws the standard "No profile provided" error before any request, re-throwing surfaces it as usual. The guard just stops the no-profile case from being reported as "could not verify".

tneum-celonis (dismissed):

src/core/feature-flag/early-access-features.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Backend feature-flag keys for early-access commands.
```
kuvia: could you clean up the excessive comments a bit? I'm not sure they're all useful, for example this one

kuvia (dismissed):

src/commands/deployment/module.ts
```diff
@@ -6,11 +6,11 @@ import { DeploymentService } from "./deployment.service";
 class Module extends IModule {
 
     public register(context: Context, configurator: Configurator): void {
-        const deploymentCommand = configurator.command("deployment").beta()
+        const deploymentCommand = configurator.command("deployment").earlyAccess()
```
ZgjimHaziri: FYI, the deployment feature is GA, so this `beta()` marker was already a leftover, and moving it off here is the right cleanup. One thing still hanging around: `docs/user-guide/deployment-commands.md` is titled "Deployment Commands (beta)", so the docs and the CLI now disagree. We know fully removing the beta tag isn't the point of this PR, so feel free to leave it for a followup PR if it's too much hassle to fold in here.

src/core/command/module-handler.ts
```diff
@@ -203,17 +196,24 @@ export class CommandConfig {
...
 
-    public beta(): CommandConfig {
-        (this.cmd as any).isBeta = true;
+    /**
+     * Marks a command as early access (internal, not shown in help). With a feature
+     * key, the command is gated at run time on that backend flag.
+     */
+    public earlyAccess(featureKey?: string): this {
+        const command = this.cmd as any
+        command.isEarlyAccess = true;
```
ZgjimHaziri: Small thing that might trip us up later: `isEarlyAccess` gets set here but nothing reads it anymore. It was only ever used by the old `CustomHelp` to render the `(beta)` label, and that's gone in this PR. The doc comment also says "not shown in help", but that isn't actually true now: these commands still appear in `--help`, just without a label. Could we pick one direction, either genuinely hide them (`cmd.hidden = true`) or drop the unused flag?

(nit: missing semicolon after `const command = this.cmd as any`.)

src/core/command/module-handler.ts
```diff
@@ -232,22 +232,30 @@ export class CommandConfig {
...
         }
-        const providedOptions = this.cmd.optsWithGlobals();
-        for (const option of this.cmd.options) {
-            const optionName = option.attributeName();
-            if ((option as any).isBeta && Object.hasOwn(providedOptions, optionName)) {
-                logger.info(chalk.yellow(`The option '${option.long}' is in beta and may change in future releases.`));
+
+        let enabled: boolean;
+        try {
+            enabled = await new FeatureFlagService(this.ctx).isEnabled(featureKey);
```
ZgjimHaziri: Bigger-picture question on the approach: this turns the CLI into a second source of truth for "is this command allowed". It reads `/api/team/features` and decides, while the command's real authorization still lives on the backend. If those two ever drift (flag present but the call still 403s, or the command works without the flag), the user is back to a confusing experience. Did we weigh the other direction, letting the command run and translating its actual 403 into the friendly "not enabled, contact support" message? That keeps a single source of truth, avoids the extra round-trip, and can't fall out of sync with the backend. This approach isn't wrong, but since nothing uses a key yet, it feels like a good moment to settle on which model we want.

src/core/command/module-handler.ts
```diff
@@ -232,22 +232,30 @@ export class CommandConfig {
...
+        } catch (error) {
+            // No profile: let the standard "no profile provided" error surface instead of masking it.
+            if (!this.ctx.profile) {
+                throw error;
             }
+            throw new GracefulError(
+                `Could not verify whether '${this.cmd.name()}' is enabled for your team. Please try again later or contact support.`
```
ZgjimHaziri: A couple of edge cases worth thinking through here:
- This is fail-closed, so if `/api/team/features` is down or slow, even a team that is entitled gets blocked. For an early-access gate that's probably an acceptable trade-off, but let's make it a deliberate one.
- An expired or invalid token comes back as a 401 and gets reported as "could not verify whether X is enabled", which points the user at the wrong problem. Might be worth letting real auth errors surface as auth errors.

src/core/command/module-handler.ts
```diff
@@ -203,17 +196,24 @@ export class CommandConfig {
         return this;
     }
 
-    public beta(): CommandConfig {
-        (this.cmd as any).isBeta = true;
+    /**
+     * Marks a command as early access (internal, not shown in help). With a feature
+     * key, the command is gated at run time on that backend flag.
+     */
+    public earlyAccess(featureKey?: string): this {
```
ZgjimHaziri: Zooming out from deployments: collapsing `beta()` and `betaOption()` into `earlyAccess()` also removes our only way to tell a user "this command (or option) is beta and may change". Deployment is GA so it doesn't need that, but I think the beta/betaOption capability (or something equivalent) is worth keeping around for future commands that genuinely are early-access or unstable. `betaOption` in particular was unused, but the ability to flag a single option that way is the kind of thing we'll likely want again rather than rebuild later. Could we keep the mechanism even if nothing uses it today?

src/core/feature-flag/feature-disabled-error.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+function extractErrorCode(error: unknown): string | undefined {
+    let text = "";
+    if (error instanceof Error) {
+        text = error.message;
+    } else if (typeof error === "string") {
+        text = error;
+    }
+    const jsonStart = text.indexOf("{");
```
kuvia, line 28: this seems unnecessary. either we return json or we don't?

ZgjimHaziri (approved):

src/core/feature-flag/feature-disabled-error.ts
admirjashari: The indexOf("{") handles cases where the error message has a prefix (e.g. "Problem creating X: Error: {json}"). A plain JSON.parse would fail there.

PR #394, opened 2026-07-09
SP-1534: add description messages where they were missing

kuvia:
#### Description

Backfill help descriptions. Legacy root commands that are used by multiple modules were added to `configureRootCommands`. General commands that are only used by one module are added to the modules themselves.

#### Relevant links

- Jira: https://celonis.atlassian.net/browse/SP-1534

#### Checklist

### DIFF
--- a/src/commands/data-pipeline/data-pool/data-pool.commands.ts
+++ b/src/commands/data-pipeline/data-pool/data-pool.commands.ts
@@ -45,7 +45,8 @@ export class DataPoolCommands {
             .description("Command to push data pools")
             .action(this.pushDataPools);
 
-        const updateCommand = configurator.command("update");
+        const updateCommand = configurator.command("update")
+            .description("Update data pool");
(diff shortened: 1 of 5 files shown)


src/commands/data-pipeline/data-pool/data-pool.commands.ts
```diff
@@ -45,7 +45,8 @@ export class DataPoolCommands {
             .description("Command to push data pools")
             .action(this.pushDataPools);
 
-        const updateCommand = configurator.command("update");
+        const updateCommand = configurator.command("update")
+            .description("Update data pool");
```
tneum-celonis, line 49: Could be a bit more elaborate, no? 😄

tneum-celonis (approved):

kuvia:
Bypassing checks because it fails for uncovered files and none of the module files are covered as of yet.

PR #396, opened 2026-07-13
SP-1297: Add download agent skill bundle command

tneum-celonis:
#### Description

Adds `content-cli asset-registry skills download --path <path> [--output <parentDir>]`, which installs a complete agent skill bundle (SKILL.md + all reference files) to the local filesystem in one command, preserving folder structure and overwriting existing files.

Builds on SP-348's single-file `skills get`: a new `listSkillFiles` API call to `GET /pacman/api/core/asset-registry/skills/{path}/files` fetches the manifest, then each entry is downloaded via the existing `getSkillFile` and written under `<output>/<skillName>/<relativePath>`. Includes a path-traversal guard so malformed manifest entries cannot escape the target directory. `FileService.writeBufferToPath` now mkdirs the file's parent (backwards-compatible for the existing basename-only callers).

#### Relevant links

- Jira: https://celonis.atlassian.net/browse/SP-1297

#### Checklist


docs/user-guide/asset-registry-commands.md
```diff
@@ -214,6 +214,38 @@ Behavior:
...
 - A missing skill or file returns a clear error such as `Problem getting SKILL.md for 'platform/missing': ...`.
 
+### Download an Entire Skill
+
+Download a complete skill bundle — its `SKILL.md` plus all reference files — into a new directory named after the skill. Folder structure inside the skill is preserved. Use this to vendor a full copy of a skill into a repo or filesystem in a single command.
+
+Download a platform skill into the current directory:
+
+```
+content-cli asset-registry skills download --path platform/content-cli-setup
```
Buqeta: We already have `content-cli asset-registry skills get --path {path of the skill} --file {specific file}`.  Adding `content-cli asset-registry skills download` seems confusing, since we can just change the behavior of `asset-registry skills get` to download the whole bundle when the `--file` option is ommitted.

cc: @ZgjimHaziri @ksalihu Would appreciate your opinions here too.

ZgjimHaziri: Agree 👍

tneum-celonis: That would break existing behavior, where by default, SKILL.md is downloaded when no file is provided

tneum-celonis: We could remove that default or we could add another flag `--all` that downloads all files, so `--all` and `--file` are mutually exclusive and if neither is provided, SKILL.md is downloaded. Wdyt?

Buqeta: We don't need to worry about breaking changes, since we added the `skills get` command 2 weeks ago ([ref](https://github.com/celonis/content-cli/pull/385)) and it's not referenced in any skills as far as I know. 

 > We could remove that default or we could add another flag --all that downloads all files, so --all and --file are mutually exclusive and if neither is provided, SKILL.md is downloaded. Wdyt?

Sounds good to me.

Buqeta (approved):

tneum-celonis:
@Buqeta I guess we shouldn't merge before the backend route is merged?

Buqeta:
> Buqeta (@Buqeta) I guess we shouldn't merge before the backend route is merged?

We can merge it, but we would need to wait for the backend PR to be merged before we can test it.

tneum-celonis:
Closed because we no longer want to expose skills publicly - see ticket for context

PR #400, opened 2026-07-28
SP-1795: remove asset-registry skills list and get

kuvia:
#### Description
remove asset-registry skills list and get

#### Relevant links
https://celonis.atlassian.net/browse/SP-1795

#### Checklist


markusherdi (dismissed):
Looks good!

Two small things:
Can you also remove the commands from docs/command-graph.html?

And there is one unused constant in tests/commands/asset-registry/asset-registry-error.spec.ts 
const SKILLS_URL = ".../asset-registry/skills";

markusherdi (approved):

PR #401, opened 2026-07-28
SP-1472: Add config branch export and import (beta)

LaberionAjvazi:
#### Description

Adds beta `config branch export` and `config branch import`. Both read and write local files by default, and mirror a branch to a Git branch one-to-one when `--gitProfile` is set. `export --all` pushes the main package and all of its branches in one go.

#### Relevant links

- Jira: https://celonis.atlassian.net/browse/SP-1472

#### Checklist


src/commands/configuration-management/branch/branch-export-import.command.service.ts
```diff
@@ -0,0 +1,294 @@
...
+        const updated = this.replaceFieldValue(raw, "key", projectKey, branchPackageKey);
+        if (updated === raw) {
+            throw new FatalError(
+                `Could not rewrite the package key in ${PACKAGE_FILE} from '${projectKey}' to '${branchPackageKey}'. Nothing was imported.`
+            );
+        }
+        fs.writeFileSync(filePath, updated, { encoding: "utf-8" });
+    }
+
+    private replaceFieldValue(content: string, field: string, from: string, to: string): string {
```
promeris, line 255: instead of a regex replacement, could we do the replacement via assignment to the JSON property that we are targeting? We are already parsing the file into a JSON structure, and feels safer to do that than the regex.

src/commands/configuration-management/branch/branch-export-import.command.service.ts
```diff
@@ -0,0 +1,294 @@
...
+                throw new FatalError("The --directory option accepts only directories.");
+            }
+            fs.cpSync(directory, workingDir, { recursive: true });
+            return workingDir;
+        }
+        if (fileService.isDirectory(file)) {
+            fs.rmSync(workingDir, { recursive: true, force: true });
+            throw new FatalError("The --file option accepts only zip files.");
+        }
+        new AdmZip(file).extractAllTo(workingDir, true);
```
promeris: I think there was some new validation to validate against large zip files. Can you double check the other cases we have in the project?

src/commands/configuration-management/branch/branch-export-import.command.service.ts
LaberionAjvazi: The regex replacement was done so that the formatting of the file is not changed due to different parser configuration. This would show as a change in github prs, so the regex replacement is intended to avoid this.

promeris (dismissed):

promeris (approved):

PR #404, opened 2026-08-07
SP-2354: Add export and import bookmarks commands

roberto-welzel:
#### Description

Add `export bookmarks` and `import bookmarks` commands that wrap the package-manager bookmark batch APIs. These are used by the content-deployment-pipeline to copy bookmarks during T2TC.

#### Relevant links

- Jira: https://celonis.atlassian.net/browse/SP-2354

#### Checklist

### DIFF
--- a/src/commands/bookmarks/bookmarks-api.ts
+++ b/src/commands/bookmarks/bookmarks-api.ts
@@ -1,6 +1,7 @@
 import { HttpClient } from "../../core/http/http-client";
 import { Context } from "../../core/command/cli-context";
 import { FatalError } from "../../core/utils/logger";
+import { BookmarksExport, BookmarksImportRequest, BookmarksImportResult } from "./bookmarks.interfaces";
 
 export class BookmarksApi {
 
@@ -10,13 +11,13 @@ export class BookmarksApi {
         this.httpClient = () => context.httpClient;
     }
 
-    public async exportBookmarks(packageKey: string): Promise<any> {
+    public async exportBookmarks(packageKey: string): Promise<BookmarksExport> {
         return this.httpClient().get(`/package-manager/api/packages/${encodeURIComponent(packageKey)}/bookmarks/export`).catch(e => {
             throw new FatalError(`Problem exporting bookmarks for package ${packageKey}: ${e}`);
         });
     }
 
-    public async importBookmarks(packageKey: string, payload: any): Promise<any> {
+    public async importBookmarks(packageKey: string, payload: BookmarksImportRequest): Promise<BookmarksImportResult> {
         return this.httpClient().post(`/package-manager/api/packages/${encodeURIComponent(packageKey)}/bookmarks/import`, payload).catch(e => {
             throw new FatalError(`Problem importing bookmarks for package ${packageKey}: ${e}`);
         });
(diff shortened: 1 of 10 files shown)


src/commands/bookmarks/bookmarks-api.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    constructor(context: Context) {
+        this.httpClient = () => context.httpClient;
+    }
+
+    public async exportBookmarks(packageKey: string): Promise<BookmarksExport> {
+        try {
+            return await this.httpClient().get(`/package-manager/api/packages/${encodeURIComponent(packageKey)}/bookmarks/export`);
+        } catch (e) {
+            throw new FatalError(`Problem exporting bookmarks for package ${packageKey}: ${e}`, { cause: e });
```
alex-probst: I looked at other places where `FatalError` is thrown and we do not set anywhere a `cause` property. Maybe it is not necessary here?

src/commands/bookmarks/bookmarks-command.service.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    constructor(context: Context) {
+        this.bookmarksApi = new BookmarksApi(context);
+    }
+
+    public async exportBookmarks(packageKey: string, file?: string): Promise<void> {
+        const exportData = await this.bookmarksApi.exportBookmarks(packageKey);
+
+        const fileName = file ?? `bookmarks-${packageKey}.json`;
+        fileService.writeToFileWithGivenName(JSON.stringify(exportData, null, 4), fileName);
```
tneum-celonis, line 18: If this write fails, is there any graceful error handling?

src/commands/bookmarks/bookmarks-command.service.ts
```diff
@@ -0,0 +1,27 @@
...
+        logger.info(FileService.fileDownloadedMessage + fileName);
+    }
+
+    public async importBookmarks(packageKey: string, file: string): Promise<void> {
+        const payload = fileService.readFileToJson(file);
```
tneum-celonis, line 23: If this read fails, is there any graceful error handling?

src/commands/bookmarks/bookmarks-api.ts
roberto-welzel: Good point — removed `{ cause: e }` to stay consistent with the rest of the codebase.

src/commands/bookmarks/bookmarks-command.service.ts
roberto-welzel: Errors from `fs.writeFileSync` propagate up to the framework's `CommandConfig.action()` handler, which catches all exceptions and logs them with `process.exitCode = 1`. This is the same pattern used by all other command services in the repo (e.g. deployment, configuration-management, t2tc, studio, etc.) — none add explicit error handling around `writeToFileWithGivenName`.

src/commands/bookmarks/bookmarks-command.service.ts
roberto-welzel: Same as above — `readFileToJson` internally checks file existence (logging a `FatalError` if missing) and then parses the JSON. Any exception propagates to the framework's error handler in `CommandConfig.action()`. This matches the pattern used by all other services that call `readFileToJson` or `readFile`.

tneum-celonis (dismissed):
Image

alex-probst (dismissed):

tneum-celonis (approved):

PR #405, opened 2026-08-07
SP-1173: add CUI marking for files the CLI writes to disk

dwoditsch:
#### Description

Every user-facing file content-cli writes to disk has to carry the team's CUI (Controlled Unclassified Information) marking. This PR adds the mechanism and wires up `list spaces --json` and `list packages --json`. The remaining writers follow in later PRs.

What a user gets depends on the team's CUI settings:

| Team setting | Artifact |
|---|---|
| marking does not apply | original filename, unchanged |
| marking applies, no categories | `Unclassified - <name>` |
| marking applies, with categories | `CUI - <name>.zip`, holding the payload plus `CUI_Cover_Sheet.pdf` |
| marking cannot be resolved | nothing written, the command fails |

This ships inert in production: the cover endpoint is allowlisted for staging only, so production keeps taking the unmarked path until the matching rule lands. Commands that only print to the console never probe the endpoint.

#### Scope: how the write is triggered

| Trigger | Commands | CUI marking |
|---|---|---|
| `--json` listings and reports | `list spaces`, `list packages` | **This PR**, together with the marking mechanism |
| `--json` listings and reports | `list assets/assignments/data-pools`, `config *`, `t2tc package list/diff`, `deployment *`, `asset-registry *` | To follow in [#407](https://github.com/celonis/content-cli/pull/407) |
| `-o, --outputToJsonFile` reports | `analyze/import action-flows`, `export data-pool`, `import data-pools`, `t2tc package import` report | To follow in [#410](https://github.com/celonis/content-cli/pull/410) |
| artifact is already an archive | `config package export --zip`, `config branch export --zip`, `t2tc package export`, `export action-flows`, `pull package` | To follow in [#411](https://github.com/celonis/content-cli/pull/411) |
| single non-archive export | `pull asset/skill/data-pool/view-bookmarks/bookmarks`, `export bookmarks` | To follow in [#412](https://github.com/celonis/content-cli/pull/412) |
| output is a directory | `config package export`, `config branch export`, `t2tc package export --unzip` | To follow in [#413](https://github.com/celonis/content-cli/pull/413): cover sheet into the directory, prefix the directory name |
| `--gitBranch` variants | `config package export`, `config branch export`, `t2tc package export` | Out of scope, nothing reaches local disk |
| no output flag | console-only listings, profile/git-profile/log files | Out of scope |

Two things were left to settle here and both are answered further up the stack: an artifact that is already a zip gets the cover sheet merged in rather than nested ([#411](https://github.com/celonis/content-cli/pull/411)), and directory output keeps the cover sheet inside the directory with the directory name prefixed.

#### Relevant links

- Jira: [SP-1173](https://celonis.atlassian.net/browse/SP-1173)
- Backend endpoint: [celonis/cloud-team#5659](https://github.com/celonis/cloud-team/pull/5659) ([VEX-3192](https://celonis.atlassian.net/browse/VEX-3192))
- Gateway allowlist (staging): [celonis/policy-guard#127](https://github.com/celonis/policy-guard/pull/127)

#### Checklist

[SP-1173]: https://celonis.atlassian.net/browse/SP-1173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Changes what files land on disk and their names for JSON listings, with an extra API call per write; scope is limited to two commands and fails closed when cover fetch fails.
> 
> **Overview**
> Adds **CUI (Controlled Unclassified Information) marking** for disk writes from `list spaces --json` and `list packages --json`, with a reusable path for later commands.
> 
> Before writing JSON, the CLI calls the team **CUI PDF cover** API. Outcomes: **no marking** (unchanged filename), **unclassified** (`Unclassified - …`), or **classified** (`CUI - ….zip` containing the JSON plus `CUI_Cover_Sheet.pdf`). If marking cannot be resolved, the command **fails** and nothing is written. **Console-only** listings skip the CUI probe.
> 
> New pieces: `CuiApi` / `CuiFileService`, `HttpClient.getStatusAndData` for status-aware GETs, and `BaseManager.findAll` awaiting async `onFindAll` handlers. Test mocks default the cover endpoint to 204 so existing tests stay unmarked unless they opt in.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ac4b2551c96329616611556cd81f133d95b98f7d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/core/utils/cui-api.ts
+++ b/src/core/utils/cui-api.ts
@@ -7,7 +7,7 @@ export interface CuiPdfCoverResponse {
     coverPage?: { pdfContent: string; encoding: string };
 }
 
-export class CuiService {
+export class CuiApi {
     private static readonly CUI_PDF_COVER_SHEET_URL = "/api/team/cui-settings/cui-pdf-cover";
 
     private static readonly STATUS_OK = 200;
@@ -21,22 +21,22 @@ export class CuiService {
     }
 
     public async getCuiPdfCover(): Promise<CuiPdfCoverResponse | null> {
-        const { status, data } = await this.httpClient().getStatusAndData(CuiService.CUI_PDF_COVER_SHEET_URL);
+        const { status, data } = await this.httpClient().getStatusAndData(CuiApi.CUI_PDF_COVER_SHEET_URL);
 
-        if (status === CuiService.STATUS_FORBIDDEN) {
+        if (status === CuiApi.STATUS_FORBIDDEN) {
             logger.debug("CUI marking does not apply, the feature flag is disabled");
             return null;
         }
 
-        if (status === CuiService.STATUS_NO_CONTENT) {
+        if (status === CuiApi.STATUS_NO_CONTENT) {
             logger.debug("CUI marking does not apply, the team has CUI disabled");
             return null;
         }
 
-        if (status !== CuiService.STATUS_OK) {
-            throw new FatalError("Problem fetching cui pdf cover");
+        if (status === CuiApi.STATUS_OK && data) {
+            return data as CuiPdfCoverResponse;
         }
 
-        return data ? (data as CuiPdfCoverResponse) : null;
+        throw new FatalError("Problem fetching cui pdf cover");
     }
 }
--- a/src/core/utils/cui-file-service.ts
+++ b/src/core/utils/cui-file-service.ts
@@ -1,0 +1,81 @@
+import * as path from "path";
+import AdmZip = require("adm-zip");
+import { Context } from "../command/cli-context";
+import { CuiPdfCoverResponse, CuiService } from "./cui-service";
+import { fileService } from "./file-service";
+import { FileConstants } from "./file.constants";
+import { FatalError } from "./logger";
+
+export class CuiFileService {
+    public static readonly COVER_SHEET_FILE_NAME = "CUI_Cover_Sheet.pdf";
+    public static readonly CLASSIFIED_PREFIX = "CUI - ";
+    public static readonly UNCLASSIFIED_PREFIX = "Unclassified - ";
+
+    private static readonly BASE64_ENCODING = "base64";
+
+    private readonly cuiService: CuiService;
+
+    constructor(context: Context) {
+        this.cuiService = new CuiService(context);
+    }
+
+    public async writeToFileWithGivenName(data: string, filename: string): Promise<string> {
+        const cover = await this.cuiService.getCuiPdfCover();
+
+        if (cover === null) {
+            fileService.writeToFileWithGivenName(data, filename);
+            return filename;
+        }
+
+        if (!this.isClassified(cover)) {
+            const unclassifiedName = this.prefixFileName(filename, CuiFileService.UNCLASSIFIED_PREFIX);
+            fileService.writeToFileWithGivenName(data, unclassifiedName);
+            return unclassifiedName;
+        }
+
+        return this.writeClassifiedArchive(filename, data, cover);
+    }
+
+    private writeClassifiedArchive(filename: string, data: string, cover: CuiPdfCoverResponse): string {
+        const zip = new AdmZip();
+        zip.addFile(path.basename(filename), Buffer.from(data, "utf-8"), "", FileConstants.DEFAULT_FILE_PERMISSIONS);
+        zip.addFile(
+            CuiFileService.COVER_SHEET_FILE_NAME,
+            this.decodeCoverPage(cover),
+            "",
+            FileConstants.DEFAULT_FILE_PERMISSIONS
+        );
+
+        const archiveName = this.buildClassifiedArchiveName(filename);
+        fileService.writeBufferToFileWithGivenName(zip.toBuffer(), archiveName);
+
+        return archiveName;
+    }
+
+    private decodeCoverPage(cover: CuiPdfCoverResponse): Buffer {
+        const coverPage = cover.coverPage;
+        if (!coverPage?.pdfContent) {
+            throw new FatalError("CUI marking applies but the response contained no cover page.");
+        }
+        if (coverPage.encoding !== CuiFileService.BASE64_ENCODING) {
+            throw new FatalError(`Unsupported CUI cover page encoding: ${coverPage.encoding}`);
+        }
+
+        return Buffer.from(coverPage.pdfContent, CuiFileService.BASE64_ENCODING);
+    }
+
+    private isClassified(cover: CuiPdfCoverResponse): boolean {
+        return (cover.resolvedCuiMarking?.categories?.length ?? 0) > 0;
+    }
+
+    private buildClassifiedArchiveName(filename: string): string {
+        const baseName = path.basename(filename);
+        const nameWithoutExtension = baseName.slice(0, baseName.length - path.extname(baseName).length);
+
+        return path.join(path.dirname(filename), `${CuiFileService.CLASSIFIED_PREFIX}${nameWithoutExtension}.zip`);
+    }
+
+    private prefixFileName(filename: string, prefix: string): string {
+        return path.join(path.dirname(filename), `${prefix}${path.basename(filename)}`);
+    }
+}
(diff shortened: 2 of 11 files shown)


src/core/utils/cui-file-service.ts
```diff
@@ -0,0 +1,81 @@
+import * as path from "node:path";
+import AdmZip = require("adm-zip");
+import { Context } from "../command/cli-context";
+import { CuiPdfCoverResponse, CuiService } from "./cui-service";
+import { fileService } from "./file-service";
+import { FileConstants } from "./file.constants";
+import { FatalError } from "./logger";
+
+export class CuiFileService {
```
nikolai-celonis, line 9: I think cui-service.ts and cui-file-service.ts is maybe a bit confusing naming. To me it looks like some other services that make API calls have an `-api` suffix.
maybe cui-pdf-api.ts or cui-api.ts ?

nikolai-celonis (dismissed):

src/core/utils/cui-file-service.ts
dwoditsch: I will tackle that in one of the follow up PRs

nikolai-celonis (approved):

alex-probst (approved):

src/core/utils/cui-api.ts
```diff
@@ -0,0 +1,42 @@
...
+    public async getCuiPdfCover(): Promise<CuiPdfCoverResponse | null> {
+        const { status, data } = await this.httpClient().getStatusAndData(CuiApi.CUI_PDF_COVER_SHEET_URL);
+
+        if (status === CuiApi.STATUS_FORBIDDEN) {
+            logger.debug("CUI marking does not apply, the feature flag is disabled");
+            return null;
+        }
+
+        if (status === CuiApi.STATUS_NO_CONTENT) {
+            logger.debug("CUI marking does not apply, the team has CUI disabled");
```
ksalihu, line 32: What is difference between the two?

src/core/utils/cui-api.ts
```diff
@@ -0,0 +1,42 @@
...
+            return null;
+        }
+
+        if (status === CuiApi.STATUS_OK && data) {
+            return data as CuiPdfCoverResponse;
+        }
```
ksalihu, line 38: Could you check the response shape here instead of just truthiness? Stubbing this endpoint with `200` + an HTML body, and with `200 {"unexpected":"shape"}`, then calling `SpaceCommandService.listSpaces(true)` on this branch, both wrote `Unclassified - <uuid>.json` and returned without error. A proxy or captive-portal interstitial answering 200 is the realistic way to hit that, and the file ends up carrying a marking nothing verified.

Requiring `resolvedCuiMarking` to be an object before returning would drop through to the `FatalError` below, matching how the empty-body case already fails closed.

src/core/utils/cui-api.ts
```diff
@@ -0,0 +1,42 @@
...
+
+        throw new FatalError("Problem fetching cui pdf cover");
```
ksalihu, line 40: Could this message tell the user what to do? Someone running `list packages --json` gets `error: Error: Problem fetching cui pdf cover`, which doesn't say the listing was not written, or that CUI marking is the reason. Something like "could not resolve CUI marking for this team - no file was written" would be actionable.

tests/core/utils/cui-file-service.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    const readFile = (filename: string): Buffer => readFileSync(resolve(process.cwd(), filename));
+
+    beforeEach(() => {
+        cuiFileService = new CuiFileService(testContext);
+    });
+
+    describe("when CUI marking does not apply", () => {
+        it("Should keep the original filename when the feature flag is disabled", async () => {
+            mockAxiosGetError(COVER_URL, 403, { errorCode: "feature-disabled" });
```
ksalihu, line 32: Could you add a case using `mockAxiosGetWithStatus(COVER_URL, 403, ...)` alongside this one? `mockAxiosGetError` rejects, but `getStatusAndData` sets `validateStatus: () => true`, so real axios resolves with the 403 and takes the other branch. I checked and it behaves correctly today, so this is coverage rather than a bug - but 403 is the path production takes until the gateway rule lands, so it is the shape worth pinning.

src/core/utils/cui-file-service.ts
```diff
@@ -0,0 +1,81 @@
...
+    private static readonly BASE64_ENCODING = "base64";
+
+    private readonly cuiApi: CuiApi;
+
+    constructor(context: Context) {
+        this.cuiApi = new CuiApi(context);
+    }
+
+    public async writeToFileWithGivenName(data: string, filename: string): Promise<string> {
+        const cover = await this.cuiApi.getCuiPdfCover();
```
ksalihu, line 23: Could the cover response be memoised per `Context` before #407 and #410-413 land? Each `CuiFileService` builds its own `CuiApi`, so once batch exports route through here every written file costs another round trip to `/api/team/cui-settings/cui-pdf-cover`. Cheaper to solve once here than in five follow-ups.

ksalihu (approved):
Approach looks sound and the staged rollout across #407-#413 makes sense to me. None of my comments block merge - fine to pick them up in the follow-up PRs.

The one I would put for consideration is only the `200`-with-an-unrecognised-body case in `cui-api.ts`. It is the only path where the feature silently produces an unverified `Unclassified` marking instead of failing closed, and it gets more surface area with every command that starts routing through `CuiFileService`. Also fine to address in followup PRs if this adds conflicts or smth

ZgjimHaziri (approved):

src/core/utils/cui-api.ts
dwoditsch: first was feature flag disabled, second was CUI setting as a feature disabled (from Admin settings). But this behavior changed with latest changes [here](https://github.com/celonis/cloud-team/pull/5693)

src/core/utils/cui-file-service.ts
dwoditsch: I thought about adding a caching to this with a field on `Context` but apparently it lives and dies with the process, so it covers one command execution and nothing beyond it. So each `content-cli ...` is a separate node process which builds a fresh `Context`. A caching across invocations in a terminal is a different mechanism which should should be evaluated carefully (like writing a temp file). Also thoughts needs to be put into TTL etc. I see the need for this, but this should be handled or discussed separately

src/core/utils/cui-api.ts
dwoditsch: not relevant anymore with the last PR as there is no cases for `content-cli` for `Unclassified` anymore

src/core/utils/cui-api.ts
dwoditsch: Good catch, addressed in #416 (stacked on #414).

All three CUI failures now share one message: `Could not resolve CUI marking. No file was written`. The specifics that used to be in the text, the missing cover page and the rejected encoding, moved to debug logs along with the response status, so `--debug` still separates the cases.

tests/core/utils/cui-file-service.spec.ts
dwoditsch: Added in #416 (stacked on #414): a case using `mockAxiosGetWithStatus(COVER_URL, 403, "")` alongside the rejecting one, so the branch production takes is pinned.

Worth noting for context: since this thread, the shared mock in `tests/utls/http-requests-mock.ts` was changed to answer the cover probe with a resolved 403 by default (commit `36861f0`), so every spec that writes a file without opting in already exercises that path. The new case makes it explicit rather than incidental.

src/core/utils/cui-file-service.ts
ksalihu: Hmm, maybe I misunderstood that part, my main concern is if e.g. when using the batch export import commands, does this call get repeated for each exported package? If so, we could "cache" this result and reuse it for the same command execution. Agreed for the cross command invocations, thats something we dont need right now and would probably just complicate things

dwoditsch: no, batch export does not repeat the call per package. It's one call per command invocation

PR #424, opened 2026-09-10
SP-878: Remove asset-registry validate command

ZgjimHaziri:
#### Description

Removes the `asset-registry validate` command.

It overlapped with `config package validate`, which already validates stored nodes and does so
across all validation layers rather than business logic only, and it can be narrowed to individual
nodes with `--nodeKeys`. The one thing it did that `config package validate` cannot is inline
validation of a configuration that is not stored yet (`--configuration`, and `-f` with inline
`nodes`) — that capability is being dropped per SP-878, so the command goes with it.

Traffic on the underlying endpoint supports this: over the last 14 days Pacman's public
`POST /api/core/asset-registry/validate/{assetType}` served 174 requests in `develop` and 5 across
all production tenants, while the validation load that matters flows through
`POST /api/core/packages/{packageKey}/validate`.

The endpoint itself stays for now. ems-frontend's ace-widget calls it for inline validation on save,
so it can only be removed once that usage is gone — a follow-up on the Pacman side once the
frontend change is rolled out.

Docs now point at `config package validate` for asset validation, and the command graph drops the
node. The integration spec keeps an `asset-registry validate` case asserting the command no longer
resolves, so a re-introduction would have to be deliberate.

#### Relevant links

- Jira: https://celonis.atlassian.net/browse/SP-878

#### Checklist


Buqeta (dismissed):

docs/user-guide/asset-registry-commands.md
```diff
@@ -72,69 +72,17 @@ Options:
...
-```json
-{
-  "assetType": "BOARD_V2",
-  "packageKey": "my-pkg",
-  "nodeKeys": ["my-view"]
-}
-```
-
-### Validate a raw configuration (`--configuration`)
+There is no `asset-registry validate` command. Validate stored assets with `config package validate`,
```
ksalihu: I think we don't need to mention the first part regarding `asset-registry validate` at all

ksalihu (dismissed):

Buqeta (approved):

